PR #419, opened 2022-09-30
Add script to create and activate DR

anajuliabit:
Code changes:

- Add new script `create-and-activate-DR.js`
- Change the statement for verifying if admin address exists on `scripts/deploy-suite.js` to: 
  ```javascript
    if (adminAddress == ethers.constants.AddressZero || !adminAddress) {
  ```
  because  `adminAddress` was undefined on my env and I got an error, so I changed to `!adminAddress` to also check undefined or null values
- Change the `tidy:scripts` and `check:scripts` to run on javascript files that belong to repository root as well


scripts/util/create-and-activate-DR.js
```diff
@@ -0,0 +1,80 @@
...
+  const signer = await ethers.getSigner(adminAddress);
+
+  // Create dispute resolver
+  let tx = await accountHandler
+    .connect(signer)
+    .createDisputeResolver(disputeResolver, disputeResolverFees, sellerAllowList);
+  await tx.wait(confirmations);
+
+  // Activate dispute resolver
+  tx = await accountHandler.connect(signer).activateDisputeResolver(disputeResolver.id);
```
zajck, line 72: `disputeResolver.id` should be read from `DisputeResolverCreated` event since value in JSON may not match the actual value.

zajck (changes requested):
I suggest adding `hasRole` check before calling `createDisputeResolver` and  `activateDisputeResolver`. It is possible that the `ADMIN` in `.env` was changed to some address that is not actual admin. In that case, this address can actually create the dispute resolver, but cannot activate it.
If then mistake in `.env` is corrected, and script is run again, it will fail already at `createDisputeResolver` step since address is used already.

You may even add optional flags `activate-only` or `create-only` so script can do only one action.


Beside that, can you please also add the short description to `docs/tasks.md`

scripts/util/create-and-activate-DR.js
anajuliabit: fixed

zajck (approved):

PR #422, opened 2022-10-04
Add Custom AuthTokenType

anajuliabit:
This PR add a `Custom` type to `AuthTokenType` enum

### DIFF
--- a/contracts/domain/BosonTypes.sol
+++ b/contracts/domain/BosonTypes.sol
@@ -65,6 +65,7 @@ contract BosonTypes {
 
     enum AuthTokenType {
         None,
+        Custom,
         Lens,
         ENS
     }
--- a/contracts/domain/BosonTypes.sol
+++ b/contracts/domain/BosonTypes.sol
@@ -65,7 +65,7 @@ contract BosonTypes {
 
     enum AuthTokenType {
         None,
-        Custom,
+        Custom, // For future use
         Lens,
         ENS
     }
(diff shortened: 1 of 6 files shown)


contracts/domain/BosonTypes.sol
```diff
@@ -65,6 +65,8 @@ contract BosonTypes {
 
     enum AuthTokenType {
         None,
+        // For future use
+        Custom,
```
cliffhall, line 69: Maybe put this on the same line like

`Custom, // For future use`

so it doesn't look like the whole list from None down is for future use

cliffhall (changes requested):
Just an update to the comment positioning.

cliffhall (approved):
LGTM! 👍

test/protocol/ConfigHandlerTest.js
```diff
@@ -1241,6 +1248,14 @@ describe("IBosonConfigHandler", function () {
           ethers.constants.AddressZero,
           "Invalid auth token contract address"
         );
+        expect(await configHandler.connect(rando).getAuthTokenContract(AuthTokenType.ENS)).to.equal(
+          ethers.constants.AddressZero,
+          "Invalid auth token contract address"
+        );
+        expect(await configHandler.connect(rando).getAuthTokenContract(AuthTokenType.Custom)).to.equal(
```
mischat, line 1255: Does it ever get here, it does seem like it would ever get here? Or perhaps I am mistaken....

mischat (commented):
the code, looks good, I have a question of the the test changes, but yeah ...

test/protocol/ConfigHandlerTest.js
hswopeams: Calling `getAuthTokenContract(AuthTokenType.Custom)` should return the zero address, so it's a valid test case. If ` != AuthTokenType.Custom` were ever removed from the `require` in `setAuthTokenContract`, this test case would fail. That's what we want.

hswopeams (approved):

PR #424, opened 2022-10-05
Add getters to BosonVoucher

anajuliabit:


mischat:
@anajuliabit what is the reasoning behind changing the type to a uint256? Everything else looks harmless, but the type change will require an update to the core components ?

anajuliabit:
@mischat the variable which stores the royalty is `uint256`
See https://github.com/bosonprotocol/boson-protocol-contracts/pull/424/files#diff-b30a98e906b874b81d7c2524442c00f39626a25b06d9bb0d0374acd811524082L29 

We weren't explicit converting to `uint256`, but the value was implicitly converted when we called `_setRoyaltyPercentage` on line https://github.com/bosonprotocol/boson-protocol-contracts/pull/424/files#diff-b30a98e906b874b81d7c2524442c00f39626a25b06d9bb0d0374acd811524082R250

Using the implicit conversion could let people think that the stored value is `uin96` when it is actually`uin256`

test/protocol/clients/BosonVoucher.js
```diff
@@ -504,6 +504,30 @@ describe("IBosonVoucher", function () {
...
+        await bosonVoucher.connect(operator).setRoyaltyPercentage(royaltyPercentage);
+
+        expect(await bosonVoucher.connect(rando).getRoyaltyPercentage()).to.equal(
+          royaltyPercentage,
+          "Invalid royalty percentage"
+        );
+      });
+    });
+
+    context("getSellerId()", function () {
```
zajck, line 522: You put the `getSellerId()` context inside `ERC2981 NFT Royalty fee` context, but it does not realy belog here. Just put it at the same level as `issueVoucher()`, `burnVoucher()` etc.

mischat:
> @mischat the variable that stores the royalty is `uint256` See https://github.com/bosonprotocol/boson-protocol-contracts/pull/424/files#diff-b30a98e906b874b81d7c2524442c00f39626a25b06d9bb0d0374acd811524082L29
> 
> We weren't explicit converting/passing a `uint256`, but the value was implicitly converted when we called `_setRoyaltyPercentage` in line https://github.com/bosonprotocol/boson-protocol-contracts/pull/424/files#diff-b30a98e906b874b81d7c2524442c00f39626a25b06d9bb0d0374acd811524082R250
> 
> Using the implicit conversion could let people think that the stored value is `uint96` when it is actually`uint256`

Thanks for that, I learnt something today. Thanks

test/protocol/clients/BosonVoucher.js
```diff
@@ -504,6 +504,30 @@ describe("IBosonVoucher", function () {
       });
     });
 
+    context("getRoyaltyPercentage()", function () {
+      it("should return the royalty fee percentage", async function () {
+        // give ownership to operator
+        await bosonVoucher.connect(protocol).transferOwnership(operator.address);
+
+        royaltyPercentage = "1000"; //15%
```
hswopeams, line 512: A small thing, but the number represents 10%, not 15%

contracts/protocol/clients/voucher/BosonVoucher.sol
```diff
@@ -49,6 +49,8 @@ contract BosonVoucher is IBosonVoucher, BeaconClientBase, OwnableUpgradeable, ER
         _setContractURI(voucherInitValues.contractURI);
 
         _setRoyaltyPercentage(voucherInitValues.royaltyPercentage);
+
+        emit VoucherInitialized(_sellerId, _royaltyPercentage, _contractURI);
```
hswopeams, line 53: We need a test case to check that this event is emitted when the voucher is initialized. Actually, we should also be checking that the OwnershipTransferred, ContractURIChanged, and RoyaltyPercentageChanged events are emitted when the voucher is initialized. It doesn't look like we check for those anywhere. This could be triggered by adding a use case that creates a seller.

hswopeams (changes requested):

contracts/protocol/clients/voucher/BosonVoucher.sol
zajck: We check that `OwnershipTransferred`, `ContractURIChanged` and `RoyaltyPercentageChanged` are emitted in
`👉 createSeller()` context.

Maybe we can check emitting `VoucherInitialized` in the same place.

test/protocol/clients/BosonVoucher.js
anajuliabit: fixed

contracts/protocol/clients/voucher/BosonVoucher.sol
```diff
@@ -123,6 +125,16 @@ contract BosonVoucher is IBosonVoucher, BeaconClientBase, OwnableUpgradeable, ER
         return exists ? offer.metadataUri : "";
     }
 
+    /**
+     * @notice Gets the seller id.
+     *
+     * @return the id for the Voucher seller
+     */
+    function getSellerId(uint256 _exchangeId) public view override returns (uint256) {
```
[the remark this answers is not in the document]
anajuliabit: fixed

contracts/protocol/clients/voucher/BosonVoucher.sol
anajuliabit: Done

hswopeams (dismissed):

cliffhall (approved):
LGTM! 👍

zajck (approved):

PR #425, opened 2022-10-05
Lock down creation of seller and dispute resolver account types

cliffhall:
## What is this?
We are locking down creation of seller and dispute resolver account types such that the caller must be the supplied admin, or in the case of the seller only, if an auth token is supplied, caller must own it.

## Why are we doing it?
In the overview section at the top of the [Omniscia report](https://omniscia.io/reports/boson-protocol-version-2/), the point out a frontrunning vulnerability for seller account creation that could lead to a loss of funds for the seller. 

>  ...we specifically refer to the creation of the various account types (buyer & seller) which allow arbitrary addresses to be specified. An unsuspecting user's "seller" account may be created with values different than the ones they specified in the creation call due to a malicious user's transaction containing the same input parameters with slight deviations and a higher transaction fee.
>
>We strongly encourage the Boson Protocol team to reconsider the permissionless-ness of the account creation functions as they may lead to fraud attempts based on different seller configurations due to the difference between what was submitted in the UI vs what was ultimately executed on-chain due to the aforementioned race condition.

We originally had built the seller account type with the constraint that its admin must be the creator, but business asked us to allow anyone to create it because a big client wanted to script the creation of all their seller's accounts from a single address that would be different from the admin address supplied to the method. We can still achieve that by simply having the single creator of the accounts immediately update each one after creation with the actual admin of their seller, so closing this hole doesn't jeopardize the business goal. And since the dispute resolver accounts have the same vulnerability, that is being given the same constraint.

## Changes
* In DisputeResolverHandlerFacet.sol,
  - In createDisputeResolver method, - Check that caller is supplied admin

* In SellerBase.sol,
  - import IERC721
  - In createSellerInternal method,
    - get the sender from msgSender()
    - moved if / else block that was checking only for unique addresses up in the method and added the admin check to both sides of the condition.
    - Check that clerk and operator are transaction sender
    - Check that admin is transaction sender or auth token owner 
* Fix unit tests to respect new seller creation logic


contracts/protocol/bases/SellerBase.sol
```diff
@@ -39,6 +41,34 @@ contract SellerBase is ProtocolBase, IBosonAccountEvents {
         // Cache protocol lookups for reference
         ProtocolLib.ProtocolLookups storage lookups = protocolLookups();
 
+        // Get message sender
+        address sender = msgSender();
+
+        // Do caller and uniqueness checks based on auth type
+        if (_seller.admin == address(0)) {
```
mischat, line 48: Should this not be explicitly checking to ensure that the _authToken isn't null ?
something like : 
if (_seller.admin == address(0) && isNotNull(_authToken))  <-- or however best to do it given the Struct ?

contracts/protocol/facets/DisputeResolverHandlerFacet.sol
```diff
@@ -50,6 +51,9 @@ contract DisputeResolverHandlerFacet is IBosonAccountEvents, ProtocolBase {
         // Cache protocol lookups for reference
         ProtocolLib.ProtocolLookups storage lookups = protocolLookups();
 
+        // Check that caller is supplied admin
+        require(_disputeResolver.admin ==  msgSender(), NOT_ADMIN);
```
mischat, line 55: Does the DR allow for AuthTokens ?

contracts/protocol/bases/SellerBase.sol
```diff
@@ -39,6 +41,34 @@ contract SellerBase is ProtocolBase, IBosonAccountEvents {
...
+            require(
+                lookups.sellerIdByOperator[_seller.admin] == 0 &&
+                    lookups.sellerIdByAdmin[_seller.admin] == 0 &&
+                    lookups.sellerIdByClerk[_seller.admin] == 0,
+                SELLER_ADDRESS_MUST_BE_UNIQUE
+            );
+        }
+
         // Check active is not set to false
         require(_seller.active, MUST_BE_ACTIVE);
```
mischat, line 73: is it not worth doing this first ? Cause if this is false we can skip all of the above ?

mischat:
i have put some comments down, thanks. It looks good, we need the tests tidied up now ... thanks cliff.

contracts/protocol/facets/DisputeResolverHandlerFacet.sol
hswopeams: No

contracts/protocol/bases/SellerBase.sol
hswopeams: This is already checked in the `require` in line 46.

contracts/protocol/bases/SellerBase.sol
```diff
@@ -20,6 +21,8 @@ contract SellerBase is ProtocolBase, IBosonAccountEvents {
      * Emits a SellerCreated event if successful.
      *
      * Reverts if:
+     * - Caller is not the supplied admin or does not own supplied auth token
+     * - Caller is not the supplied operator and clerk
```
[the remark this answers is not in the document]
zajck: Why the check for treasury isn't done:
- frontrunner cannot create seller or DR with fields {admin: victim, operator: victim, clerk: victim, treasury: attacker}. It will fail, since new check will detect that msg.sender (==atacker) does not match the victim's address
- treasury cannot perform any action in the protocol

contracts/protocol/bases/SellerBase.sol
```diff
@@ -49,6 +49,40 @@ contract SellerBase is ProtocolBase, IBosonAccountEvents {
...
         );
 
+        // Cache protocol lookups for reference
+        ProtocolLib.ProtocolLookups storage lookups = protocolLookups();
+
+        // Get message sender
+        address sender = msgSender();
+
+        // Check that caller is the supplied operator and clerk
+        require(_seller.operator == sender && _seller.clerk == sender, NOT_OPERATOR_AND_CLERK);
```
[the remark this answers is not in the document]
zajck: Why the check for treasury isn't done:
- frontrunner cannot create seller or DR with fields {admin: victim, operator: victim, clerk: victim, treasury: attacker}. It will fail, since new check will detect that msg.sender (==atacker) does not match the victim's address
- treasury cannot perform any action in the protocol

hswopeams (changes requested):
Need to add a revert reason to NatSpec in one of the contracts
Question about uniqueness checks in createSellerInternal

contracts/protocol/facets/DisputeResolverHandlerFacet.sol
mischat: great, thanks for clarifying

contracts/protocol/bases/SellerBase.sol
```diff
@@ -49,6 +49,40 @@ contract SellerBase is ProtocolBase, IBosonAccountEvents {
...
+            // Check that the admin address is unique to one seller id, across all roles
+            require(
+                lookups.sellerIdByOperator[_seller.admin] == 0 &&
+                    lookups.sellerIdByAdmin[_seller.admin] == 0 &&
+                    lookups.sellerIdByClerk[_seller.admin] == 0,
+                SELLER_ADDRESS_MUST_BE_UNIQUE
+            );
+        }
+
         // Check that the addresses are unique to one seller id, across all roles. These addresses should always be checked. Treasury is not checked
```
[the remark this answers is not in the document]
zajck: True, they are not needed. I removed them.

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -28,6 +28,7 @@ contract SellerHandlerFacet is SellerBase {
      * Emits a SellerCreated event if successful.
      *
      * Reverts if:
+     * - Caller is not the supplied admin or does not own supplied auth token
      * - The sellers region of protocol is paused
```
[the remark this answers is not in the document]
zajck: Added

hswopeams (dismissed):

contracts/protocol/bases/SellerBase.sol
```diff
@@ -49,33 +49,36 @@ contract SellerBase is ProtocolBase, IBosonAccountEvents {
...
-        mapping(address => uint256) storage sellerIdByClerk = lookups.sellerIdByClerk;
-        require(
-            sellerIdByOperator[_seller.operator] == 0 &&
-                sellerIdByOperator[_seller.clerk] == 0 &&
-                sellerIdByAdmin[_seller.operator] == 0 &&
-                sellerIdByAdmin[_seller.clerk] == 0 &&
-                sellerIdByClerk[_seller.operator] == 0 &&
-                sellerIdByClerk[_seller.clerk] == 0,
-            SELLER_ADDRESS_MUST_BE_UNIQUE
-        );
```
anajuliabit, line 64: I believe we still have to check:

```javascript
require(sellerIIdByClerk[sender] = 0 && sellerIdByOperator[sender] = 0
```

otherwise, we won't be validating when is auth type

test/protocol/SellerHandlerTest.js
```diff
@@ -552,99 +581,140 @@ describe("SellerHandler", function () {
...
+          // Attempt to Create a seller with non-unique admin, expecting revert
           await expect(
-            accountHandler.connect(rando).createSeller(seller, emptyAuthToken, voucherInitValues)
-          ).to.revertedWith(RevertReasons.INVALID_ADDRESS);
+            accountHandler.connect(admin).createSeller(seller, emptyAuthToken, voucherInitValues)
+          ).to.revertedWith(RevertReasons.SELLER_ADDRESS_MUST_BE_UNIQUE);
 
-          seller.clerk = clerk.address;
-          seller.treasury = ethers.constants.AddressZero;
+          // Update seller operator
```
anajuliabit, line 609:
```suggestion
          // Update seller clerk
```

test/protocol/SellerHandlerTest.js
```diff
@@ -552,99 +581,140 @@ describe("SellerHandler", function () {
...
           await expect(
-            accountHandler.connect(rando).createSeller(seller, emptyAuthToken, voucherInitValues)
+            accountHandler.connect(authTokenOwner).createSeller(seller2, authToken, voucherInitValues)
           ).to.revertedWith(RevertReasons.SELLER_ADDRESS_MUST_BE_UNIQUE);
 
-          //Set seller 2's clerk address to seller 1's admin address
-          seller.admin = other1.address;
-          seller.operator = other2.address;
-          seller.clerk = admin.address;
+          // Update the seller, so admin matches authTokenOwner
```
anajuliabit, line 703:
```suggestion
          // Update the seller admin and clerk, so admin matches authTokenOwner
```

test/protocol/SellerHandlerTest.js
```diff
@@ -552,99 +581,140 @@ describe("SellerHandler", function () {
...
-          //Set seller 2's admin address to seller 1's operator address
-          seller.admin = operator.address;
-          seller.operator = other2.address;
-          seller.clerk = other3.address;
+          // Update the seller, so operator matches authTokenOwner
+          seller.operator = authTokenOwner.address;
+          await accountHandler.connect(rando).updateSeller(seller, emptyAuthToken);
 
-          // Attempt to Create a seller with non-unique operator, expecting revert
+          // // Attempt to Create a seller with non-unique operator, expecting revert
```
anajuliabit, line 688:
```suggestion
          // Attempt to Create a seller with non-unique operator, expecting revert
```

anajuliabit (changes requested):
Just comments typos. I'm happy to make changes in case you are already done for today @zajck

hswopeams (approved):

anajuliabit (approved):

hswopeams (approved):

PR #426, opened 2022-10-06
Change nonce check to use signer instead of msg.sender

anajuliabit:
This PR changes the nonce check to use signer instead of `msg.sender` (relayer)


cliffhall (approved):
LGTM! 👍

hswopeams (approved):

PR #428, opened 2022-10-07
Update DR creation script

zajck:
This is branched from #425 which should be approved before this.

#425 introduces changes, which brake the creation of a DR with the script. I updated it, so now the DR is first created by the protocol admin (with admin, operator and clerk matching its address). Then if supplied DR is different, it performs updateDisputeResolver in the next step.

If we wanted to enable 1 step creation, we'd need to supply also DR's private key, which may not be available. I can add that as well if you think it's necessary.


hswopeams (changes requested):

anajuliabit (approved):
LGTM

docs/tasks.md
```diff
@@ -128,12 +128,17 @@ Script will create and/or activate a dispute resolver
...
     }
   ```
 - `network`: Network to run the script
 - `activate-only (optional)`: Optional flag to only activate the Dispute Resolver
 - `create-only (optional)`: Optional flag to only create the Dispute Resolver
 
+`privateKey` represents the private key that will create a dispute resolver. If it is not specified, the protocol admin account will be used (same account that later activate the dispute resolver and is specified in `.env`).
+If `operator`, `admin` and `clerk` matches the address, corresponding to `privateKey`, dispute resolver is simply created and activated.
+If any of `operator`, `admin` and `clerk` differs from the address, corresponding to `privateKey`, dispute resolver is created in two steps. Firstly, a dispute resolver with `operator`, `admin` and `clerk` set to address, corresponding to `privateKey` is created and then in the second step dispute resolver is updated with addresses from JSON file.
+
```
anajuliabit, line 141:
```suggestion
- `network`: Network to run the script
- `activate-only (optional)`: Optional flag to only activate the Dispute Resolver
- `create-only (optional)`: Optional flag to only create the Dispute Resolver
- `privateKey` represents the private key that will create a dispute resolver. If it is not specified, the protocol admin account will be used (same account that later activate the dispute resolver and is specified in `.env`).
  If `operator`, `admin` and `clerk` matches the address, corresponding to `privateKey`, dispute resolver is simply created and activated.
  If any of `operator`, `admin` and `clerk` differs from the address, corresponding to `privateKey`, dispute resolver is created in two steps. Firstly, a dispute resolver with `operator`, `admin` and `clerk` set to address, corresponding to `privateKey` is created and then in the second step dispute resolver is updated with addresses from JSON file.

```

anajuliabit (changes requested):
Just one thing about markdown format

cliffhall (approved):
LGTM! 👍

scripts/util/create-and-activate-DR.js
```diff
@@ -88,11 +88,33 @@ const createAndActivateDR = async (path, createOnly, activateOnly) => {
   let tx, receipt;
   // Create dispute resolver
   if (!activateOnly) {
+    // create dispute resolver with callers account
+    let adminDisputeResolver = { ...disputeResolver };
+    adminDisputeResolver.admin = adminAddress;
+    adminDisputeResolver.operator = adminAddress;
+    adminDisputeResolver.clerk = adminAddress;
```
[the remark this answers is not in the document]
cliffhall: It would complicate the script right now because you'd have to get the DR admin funded and have that account do the creation, since all the addresses have to be the same as the caller. 

Having the deployer/admin do it is a convenience for now, but when we get to the "opt-in to update" functionality, this script will have to fund all those DR addresses, and we will need the DR admin do the create with all addresses set to it, followed by an upgrade and confirmations from all its proposed new addresses.

So for the moment, I think this is OK, because we'll have to change very soon anyhow.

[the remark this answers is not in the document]
hswopeams: "this script will have to fund all those DR addresses" - The script doesn't have to do any funding. The address owner does.

"ollowed by an upgrade and confirmations from all its proposed new addresses." - The script won't need to do the update. The script only needs to create the DR using the DR admin address (which is the same as the other addresses). Then the protocol ADMIN calls activate. There's no update involved.

docs/tasks.md
zajck: I changed it, but slightly than suggested, since I don't want that it would look like `privateKey` is one of the arguments, similar to `path`, `network` etc

anajuliabit (dismissed):

docs/tasks.md
```diff
@@ -128,12 +128,19 @@ Script will create and/or activate a dispute resolver
...
   ```
 - `network`: Network to run the script
 - `activate-only (optional)`: Optional flag to only activate the Dispute Resolver
 - `create-only (optional)`: Optional flag to only create the Dispute Resolver
 
+
+Note about the field `privateKey` in JSON file:
+- `privateKey` represents the hex encoded private key that will create a dispute resolver. If it is not specified, the protocol admin account will be used (same account that later activate the dispute resolver and is specified in `.env`).
+- If all `operator`, `admin` and `clerk` match the address, corresponding to `privateKey`, dispute resolver is simply created and activated.
+- If any of `operator`, `admin` or `clerk` differs from the address, corresponding to `privateKey`, dispute resolver is created in two steps. Firstly, a dispute resolver with `operator`, `admin` and `clerk` set to address, corresponding to `privateKey` is created and then in the second step dispute resolver is updated with addresses from JSON file.
```
hswopeams, line 142: Won't this fail given the changes we are making to createDisputeResolver?

hswopeams (dismissed):

hswopeams (approved):

anajuliabit (approved):

PR #429, opened 2022-10-07
 Add opt-in feature when updating seller admin, clerk, operator and auth token

anajuliabit:
This PR add an opt-in feature when updating seller admin, clerk, operator, or auth token to prevent front-running. 

New update seller steps (except for treasury which is updated automatically):
1. Seller requests to update a field (admin, operator, clerk, or auth token)
2. Event `SellerUpdatePending` is emitted
3. Addresses owner's calls `optInToSellerUpdate` passing the `sellerId` and the fields in which they are approving the update,  
   ```javascript
        optInToSellerUpdate("1", [SellerFields.Admin, SellerFields.Operator, SellerFields.AuthToken])
   ```
4. Event `SellerUpdateApplied emitted`


contracts/interfaces/events/IBosonAccountEvents.sol
```diff
@@ -22,6 +22,11 @@ interface IBosonAccountEvents {
         BosonTypes.AuthToken authToken,
         address indexed executedBy
     );
+    event SellerUpdateAddressesRequested(
+        uint256 indexed sellerId,
+        BosonTypes.SellerUpdate seller,
+        address indexed executedBy
```
mischat, line 28: This is a note to myself, it seems like this could be performed by the authToken too, i guess that should be a part of this struct ...

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -102,33 +102,155 @@ contract SellerHandlerFacet is SellerBase {
...
+            uint256 check2 = lookups.sellerIdByClerk[_seller.clerk];
+            uint256 check3 = lookups.sellerIdByAdmin[_seller.clerk];
+
+             require(
+                (check1 == 0 || check1 == _seller.id) &&
+                    (check2 == 0 || check2 == _seller.id) &&
+                    (check3 == 0 || check3 == _seller.id),
+                SELLER_ADDRESS_MUST_BE_UNIQUE
+            );
+        }
```
mischat, line 250: Could this not be written as something like : 
```
uint256 check = 0;
if (_seller.clerk != address(0)) {
    check += lookups.sellerIdByOperator[_seller.clerk];
    check += lookups.sellerIdByClerk[_seller.clerk];
    check += lookups.sellerIdByAdmin[_seller.clerk];
}
require(check == 0, SELLER_ADDRESS_MUST_BE_UNIQUE);
```
Seems easier to read, i find the require statement rather hard to process / validate.

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -102,33 +102,155 @@ contract SellerHandlerFacet is SellerBase {
...
+        SellerUpdate storage _sellerPendingUpdate = lookups.sellerPendingUpdates[_sellerId];
+
+        require(_sellerPendingUpdate.admin !== address(0) || _sellerPendingUpdate.operator !== address(0) || _sellerPendingUpdate.clerk !== address(0), NO_PENDING_ADDRESSES_UPDATE );
+        preSellerUpdateChecks(_sellerId, _sellerPendingUpdate);
+
+        // Get storage location for seller
+        (, Seller storage seller) = fetchSeller(_sellerId);
+
+        // Approve operator address
+        if (sellerPendingUpdate.operator == sender) {
```
mischat, line 171: s/sellerPendingUpdate.admin/_sellerPendingUpdate.admin/

anajuliabit: fixed

contracts/protocol/facets/SellerHandlerFacet.sol
anajuliabit: can't, because we have to compare each check against the sellerId  ```check1 == _sellerId```, ```check2 == _sellerId```, etc.

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -102,65 +102,168 @@ contract SellerHandlerFacet is SellerBase {
...
+                // Update seller id by operator mapping
+                delete lookups.sellerIdByOperator[seller.operator];
+                lookups.sellerIdByOperator[sender] = _sellerId;
+            }
+
+            // Delete pending update operator
+            _sellerPendingUpdate.operator = address(0);
+        }
+
+        // Approva admin address
```
hswopeams, line 189:
```suggestion
        // Approve admin address
```

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -102,65 +102,168 @@ contract SellerHandlerFacet is SellerBase {
...
+
+                // Update seller id by clerk mapping
+                delete lookups.sellerIdByClerk[seller.clerk];
+                lookups.sellerIdByClerk[sender] = _sellerId;
+            }
+
+            // Delete pending update clerk
+            _sellerPendingUpdate.clerk = address(0);
+        }
+
```
hswopeams, line 218: I think if all three updateable addresses are address(0), delete `sellerPendingUpdates[sellerId]` should be called.

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -102,65 +102,168 @@ contract SellerHandlerFacet is SellerBase {
...
-        }
+        SellerRoleUpdate memory sellerUpdate;
+        sellerUpdate.operator = seller.operator;
+        sellerUpdate.clerk = seller.clerk;
+        sellerUpdate.admin = seller.admin;
+
+        preSellerUpdateChecks(_seller.id, sellerUpdate);
 
         // Admin address or AuthToken data must be present in parameters. A seller can have one or the other. Check passed in parameters
         if (_seller.admin == address(0)) {
```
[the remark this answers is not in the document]
anajuliabit: done

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -102,65 +102,168 @@ contract SellerHandlerFacet is SellerBase {
...
+
+            // Delete pending update clerk
+            _sellerPendingUpdate.clerk = address(0);
+        }
+
+        // Notify watchers of state change
+        emit SellerUpdateRoleApproved(_sellerId, seller, sender);
+    }
+
+    function preSellerUpdateChecks(uint256 _sellerId, SellerRoleUpdate memory _seller) internal view {
```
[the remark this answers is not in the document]
anajuliabit: done

contracts/protocol/facets/SellerHandlerFacet.sol
anajuliabit: done

contracts/protocol/facets/SellerHandlerFacet.sol
anajuliabit: We are reverting the transaction if all three updateable addresses are 0

```solidity
require(
            _sellerPendingUpdate.admin != address(0) ||
                _sellerPendingUpdate.operator != address(0) ||
                _sellerPendingUpdate.clerk != address(0),
            NO_PENDING_SELLER_ROLE_UPDATE
        );

```

I can remove this require and add your suggestion.
What do you prefer?

contracts/interfaces/events/IBosonAccountEvents.sol
anajuliabit: @hswopeams I remove the `UpdateSeller` struct, and now I'm using the `Seller` struct itself. 

I don't think its necessary the pending feature for the AuthToken update

zajck (changes requested):
`lookups.sellerPendingUpdates[_seller.id]` should be deleted right after `sellerPreUpdateChecks` in `updateSeller` so we always start with clean struct.

Otherwise the following can happen:
1. I have a seller with `{admin: address1, operator: address1, clerk: address1, treasury: address1}`
2. I want to update the clerk to `address2`, but I make mistake and submit `{admin: address1, operator: address2, clerk: address1, treasury: address1}`. Now operator in `sellerPendingUpdates` is set to `address2`.
3. I notice the mistake and submit a correct update request {admin: address1, operator: address1, clerk: address2, treasury: address1}`. Now clerk in `sellerPendingUpdates` is set to `address2`. But operator in `sellerPendingUpdates` was not set to 0 during this call and is still set to `address2`. 
4. At the same time emitted event `SellerUpdateRolesRequested` will show the struct I passed in the second tx and everything will look ok to me.
5. I call `approveSellerUpdate` with address2 and both `operator` and `clerk` are updated to address2 which is not what I want.

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -94,44 +94,176 @@ contract SellerHandlerFacet is SellerBase {
...
         } else {
             require(seller.admin == sender, NOT_ADMIN);
         }
 
-        // Check that the passed in addresses are unique to one seller id across all roles -- not used or are used by this seller id.
-        // Checking this seller id is necessary because one or more addresses may not change
-        {
+        sellerPreUpdateChecks(_seller);
+
+        bool needsApproval = false;
```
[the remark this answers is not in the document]
anajuliabit: fixed

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -94,44 +94,176 @@ contract SellerHandlerFacet is SellerBase {
...
+
+                // Transfer ownership of NFT voucher to new operator
+                IBosonVoucher(lookups.cloneAddress[seller.id]).transferOwnership(seller.operator);
+
+                // Store new seller id by operator mapping
+                lookups.sellerIdByOperator[sender] = _sellerId;
+            }
+
+            // Delete pending update operator
+            _sellerPendingUpdate.operator = address(0);
```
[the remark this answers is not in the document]
anajuliabit: done

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -94,44 +94,176 @@ contract SellerHandlerFacet is SellerBase {
...
+            delete lookups.sellerIdByAuthToken[authToken.tokenType][authToken.tokenId];
+
+            // Store seller by auth token reference
+            lookups.sellerIdByAuthToken[_authToken.tokenType][_authToken.tokenId] = _seller.id;
+
+            // Remove previous admin address if it exists
+            delete lookups.sellerIdByAdmin[seller.admin];
+            delete seller.admin;
+        } else {
+            require(_seller.admin != address(0), INVALID_ADDRESS);
```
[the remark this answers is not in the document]
anajuliabit: done

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -94,44 +94,176 @@ contract SellerHandlerFacet is SellerBase {
...
+                lookups.sellerIdByOperator[sender] = _sellerId;
+            }
+
+            // Delete pending update operator
+            _sellerPendingUpdate.operator = address(0);
+        }
+
+        // Approve admin address
+        if (_sellerPendingUpdate.admin == sender) {
+            if (seller.admin != _sellerPendingUpdate.admin) {
```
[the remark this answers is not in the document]
anajuliabit: not sure if I understood. If we change for `_sellerPendingUpdate.admin != address(0)` this will be a different validation. We can't remove `_sellerPendingUpdate.admin == sender` because this is what check if the caller is the correct address, and why update even when the current admin is equal to new admin?

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -94,44 +94,176 @@ contract SellerHandlerFacet is SellerBase {
...
+
+            // Delete pending update clerk
+            _sellerPendingUpdate.clerk = address(0);
+        }
+
+        // Notify watchers of state change
+        emit SellerUpdateRolesApproved(_sellerId, seller, sender);
+    }
+
+    function sellerPreUpdateChecks(Seller memory _seller) internal view {
```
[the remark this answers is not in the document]
anajuliabit: refactored

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -94,44 +94,176 @@ contract SellerHandlerFacet is SellerBase {
...
+        if (seller.treasury != _seller.treasury) {
+            require(_seller.treasury != address(0), INVALID_ADDRESS);
+            // Update treasury
+            seller.treasury = _seller.treasury;
+        }
+
+        if (_seller.operator != seller.operator) {
+            require(_seller.operator != address(0), INVALID_ADDRESS);
+            // Operator address owner must approve the update for prevent front-running
+            lookups.sellerPendingUpdates[seller.id].operator = _seller.operator;
```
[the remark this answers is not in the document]
anajuliabit: done

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -94,44 +94,176 @@ contract SellerHandlerFacet is SellerBase {
...
+
+        if (_seller.clerk != seller.clerk) {
+            require(_seller.clerk != address(0), INVALID_ADDRESS);
+            // Clerk address owner must approve the update for prevent front-running
+            lookups.sellerPendingUpdates[seller.id].clerk = _seller.clerk;
+            needsApproval = true;
+        }
+
+        if (needsApproval) {
+            lookups.sellerPendingUpdates[seller.id].id = _seller.id;
```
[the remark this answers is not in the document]
anajuliabit: done

cliffhall (changes requested):
Some requests

contracts/protocol/facets/SellerHandlerFacet.sol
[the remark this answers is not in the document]
zajck: Yes, what I wrote wasn't accurate. What I meant was, can it really happen that `_sellerPendingUpdate.admin` and `seller.admin` are the same?

I thought that [this check](https://github.com/bosonprotocol/boson-protocol-contracts/blob/96693a2c6ba18285c0b3296e8631d850cba05830/contracts/protocol/facets/SellerHandlerFacet.sol#L133)
prevents it.

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -94,73 +94,178 @@ contract SellerHandlerFacet is SellerBase {
...
+        if (_sellerPendingUpdate.operator == sender) {
+            // If operator changed, update the operator and transfer the ownership of NFT voucher
+            if (seller.operator != _sellerPendingUpdate.operator) {
+                delete lookups.sellerIdByOperator[seller.operator];
+
+                // Update operator
+                seller.operator = _sellerPendingUpdate.operator;
+
+                // Transfer ownership of NFT voucher to new operator
+                IBosonVoucher(lookups.cloneAddress[seller.id]).transferOwnership(seller.operator);
```
zajck, line 201:
```suggestion
                IBosonVoucher(lookups.cloneAddress[_sellerId]).transferOwnership(seller.operator);
```

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -94,73 +94,178 @@ contract SellerHandlerFacet is SellerBase {
...
+                lookups.sellerIdByAdmin[sender] = _sellerId;
+            }
 
-        // If operator changed, transfer the ownership of NFT voucher
-        if (oldSellerOperator != _seller.operator) {
-            IBosonVoucher(lookups.cloneAddress[seller.id]).transferOwnership(_seller.operator);
+            // Delete pending update admin
+            delete _sellerPendingUpdate.admin;
+            // Delete auth token for seller id if it exists
+            delete protocolEntities().authTokens[seller.id];
```
zajck, line 227:
```suggestion
            delete protocolEntities().authTokens[_sellerId];
```

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -94,73 +94,178 @@ contract SellerHandlerFacet is SellerBase {
...
+        require(
+            _sellerPendingUpdate.admin != address(0) ||
+                _sellerPendingUpdate.operator != address(0) ||
+                _sellerPendingUpdate.clerk != address(0),
+            NO_PENDING_SELLER_ROLE_UPDATE
+        );
+
+        sellerPreUpdateChecks(_sellerId, _sellerPendingUpdate.admin, lookups);
+        sellerPreUpdateChecks(_sellerId, _sellerPendingUpdate.operator, lookups);
+        sellerPreUpdateChecks(_sellerId, _sellerPendingUpdate.clerk, lookups);
```
zajck, line 186: What if you do here just one `sellerPreUpdateChecks(_sellerId, sender, lookups);`?

For pending updates where role address does not match the sender, it doesn't even matter if they are unique or not, since they will not be updated.
And for pending updates that matches sender it's sufficient to do the check once, since the address is the same.
It seems to me, that it should be ok.

zajck (changes requested):

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -94,73 +94,178 @@ contract SellerHandlerFacet is SellerBase {
...
             require(check == 0 || check == _seller.id, AUTH_TOKEN_MUST_BE_UNIQUE);
+
+            delete lookups.sellerIdByAuthToken[authToken.tokenType][authToken.tokenId];
+
+            // Store auth token
+            authToken.tokenId = _authToken.tokenId;
+            authToken.tokenType = _authToken.tokenType;
+
+            // Store seller by auth token reference
+            lookups.sellerIdByAuthToken[_authToken.tokenType][_authToken.tokenId] = _seller.id;
```
hswopeams, line 127: When we first implemented `createSeller`, we did not check that any addresses == caller or that the caller owns the AuthToken. We therefore also didn't check in `updateSeller` that  the new passed in AuthToken was owned by the caller. We only checked that the caller was owner of the STORED AuthToken to validate if the caller can call the function.  Now that we do check that the caller owns the AuthToken in `createSeller`, it seems to me we should also check that the caller owns the new passed in AuthToken in `updateSeller`, especially if there is no approval step for changing the AuthToken.

@zajck @anajuliabit @cliffhall Do you agree?

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -94,73 +94,178 @@ contract SellerHandlerFacet is SellerBase {
...
-            );
+            if (_seller.admin != seller.admin) {
+                require(_seller.admin != address(0), INVALID_ADDRESS);
+                // If admin address exists, admin address owner must approve the update for prevent front-running
+                lookups.sellerPendingUpdates[_seller.id].admin = _seller.admin;
+                needsApproval = true;
+            }
+        }
+
+        if (seller.treasury != _seller.treasury) {
```
hswopeams, line 141: Is there a reason for `seller` and `_seller` to be in a different order in this `If `clause?

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -94,73 +94,178 @@ contract SellerHandlerFacet is SellerBase {
...
-            uint256 check3 = lookups.sellerIdByClerk[_seller.admin];
-            require(
-                (check1 == 0 || check1 == _seller.id) &&
-                    (check2 == 0 || check2 == _seller.id) &&
-                    (check3 == 0 || check3 == _seller.id),
-                SELLER_ADDRESS_MUST_BE_UNIQUE
-            );
+            if (_seller.admin != seller.admin) {
+                require(_seller.admin != address(0), INVALID_ADDRESS);
+                // If admin address exists, admin address owner must approve the update for prevent front-running
```
hswopeams, line 135:
```suggestion
                // If admin address exists, admin address owner must approve the update to prevent front-running
```

contracts/protocol/facets/SellerHandlerFacet.sol
hswopeams: I think ` sender` or ` _sellerPendingUpdate.operator` could also be used as the parameter to `transerOwnership` because passing `seller.operator` causes an unnecessary SLOAD. To me it also makes it clearer what is happening.
Whichever variable us used should also be used in line 204 as the key. Right now there are several different variables being used that should have the same value.

hswopeams (changes requested):

contracts/protocol/facets/SellerHandlerFacet.sol
zajck: Well that would mean that `caller = ownerOf(storedAuthToken)` and `caller = ownerOf(newAuthToken)` simultaneously, which forces `ownerOf(storedAuthToken) = ownerOf(newAuthToken)` 

But what if you are trying to change AuthToken because you actually want to transfer admin rights to some other address?
Maybe we need also approval step for changing the auth token :thinking:

contracts/protocol/facets/SellerHandlerFacet.sol
zajck: `_sellerPendingUpdate.operator` also reads from storage, so I suppose we should just use `sender`.

contracts/interfaces/events/IBosonAccountEvents.sol
```diff
@@ -22,6 +22,8 @@ interface IBosonAccountEvents {
         BosonTypes.AuthToken authToken,
         address indexed executedBy
     );
+    event SellerUpdateRolesRequested(uint256 indexed accountId, BosonTypes.Seller seller, address indexed executedBy);
```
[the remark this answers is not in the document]
anajuliabit: we decide to use `SellerUpdateApplied` and `SellerUpdatePending`

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -94,73 +94,179 @@ contract SellerHandlerFacet is SellerBase {
...
+        if (needsApproval) {
+            lookups.sellerPendingUpdates[_seller.id].id = _seller.id;
+            emit SellerUpdateRolesRequested(_seller.id, _seller, sender);
+        }
+
+        // Notify watchers of state change
+        emit SellerUpdated(_seller.id, seller, _authToken, sender);
+    }
+
+    function approveSellerUpdate(uint256 _sellerId) external {
```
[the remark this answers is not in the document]
anajuliabit: we decide to use `optInToSellerUpdate`

contracts/protocol/facets/SellerHandlerFacet.sol
anajuliabit: fixed on both places

contracts/protocol/facets/SellerHandlerFacet.sol
anajuliabit: fixed

contracts/protocol/facets/SellerHandlerFacet.sol
anajuliabit: nice catch, done

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -94,73 +94,178 @@ contract SellerHandlerFacet is SellerBase {
...
-        emit SellerUpdated(_seller.id, _seller, _authToken, sender);
+        emit SellerUpdateRolesApproved(_sellerId, seller, sender);
+    }
+
+    function sellerPreUpdateChecks(
+        uint256 _sellerId,
+        address _role,
+        ProtocolLib.ProtocolLookups storage _lookups
+    ) internal view {
+        // Check that the role is unique to one seller id across all roles -- not used or are used by this seller id.
```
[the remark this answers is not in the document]
anajuliabit: fixed

contracts/protocol/facets/SellerHandlerFacet.sol
anajuliabit: fixed

contracts/protocol/facets/SellerHandlerFacet.sol
anajuliabit: @zajck good point, I agree. I think we also need approval for changing the auth token. What do you think @cliffhall @mischat @hswopeams ?

contracts/protocol/facets/SellerHandlerFacet.sol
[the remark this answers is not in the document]
anajuliabit: make sense, fixed

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -94,73 +94,176 @@ contract SellerHandlerFacet is SellerBase {
...
+            NO_PENDING_SELLER_ROLE_UPDATE
+        );
+
+        preUpdateSellerCheck(_sellerId, sender, lookups);
+
+        // Get storage location for seller
+        (, Seller storage seller, AuthToken storage authToken) = fetchSeller(_sellerId);
+
+        // Approve operator address
+        if (_sellerPendingUpdate.operator == sender && _sellerPendingUpdate.operator != address(0)) {
```
zajck, line 197: Me again on this line :)

 If `_sellerPendingUpdate.operator == sender` is true, then well, the second part `_sellerPendingUpdate.operator != address(0)` is also always true (since `sender != address(0)`).
 And if `_sellerPendingUpdate.operator == sender` is false, then second part is not even checked.
 
 So ultimately what you need is only the first part of the check before `&&`.

Sorry if my first comment about this line misled you.

zajck (changes requested):

contracts/protocol/facets/SellerHandlerFacet.sol
cliffhall: Yes, I agree. The owner of the new auth token should confirm their opt-in, just as if an address had been supplied.

contracts/protocol/facets/SellerHandlerFacet.sol
anajuliabit: done

test/protocol/SellerHandlerTest.js
```diff
@@ -1967,5 +2343,257 @@ describe("SellerHandler", function () {
...
+          sellerStruct = seller.toStruct();
+
+          await accountHandler.connect(admin).updateSeller(seller, emptyAuthToken);
+
+          await expect(
+            accountHandler.connect(other2).optInToSellerUpdate(seller.id, [SellerFields.Operator])
+          ).to.revertedWith(RevertReasons.UNAUTHORIZED_CALLER_UPDATE);
+        });
+
+        it("Should revert if the caller is not the auth token owner", async function () {
```
hswopeams, line 2573:
```suggestion
        it("Should revert if the caller is not the new auth token owner", async function () {
```

test/protocol/SellerHandlerTest.js
```diff
@@ -1967,5 +2343,257 @@ describe("SellerHandler", function () {
...
+            sellerStruct,
+            pendingSellerUpdateStruct,
+            authTokenStruct,
+            emptyAuthTokenStruct,
+            authTokenOwner.address
+          );
+      });
+
+      context("💔 Revert Reasons", async function () {
+        it("Should revert if there are no pending updates", async function () {
```
hswopeams, line 2482: The use of "Should revert if" in these test cases is inconsistent. It looks like in most places in this file we do not include this text at the begging of the revert reason test name.

hswopeams (changes requested):
It looks pretty good. There are just a few more changes needed.

We also need to look at storeSeller again to see if it should be refactored since it is no longer being used by both createSeller and updateSeller. That could be done in another PR though.

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -56,7 +56,9 @@ contract SellerHandlerFacet is SellerBase {
      *         All other fields should be filled, even those staying the same.
      * @dev    Active flag passed in by caller will be ignored. The value from storage will be used.
      *
-     * Emits a SellerUpdated event if successful.
+     * Emits a SellerUpdateApplied event if the seller has changed the treasury.
+     * Emits a SellerUpdatePending event if the seller has requested an update for admin, clerk, operator, or auth token.
+     * Addresses owner of new values for admin, clerk, operator or auth token need to opt-in to update.
```
cliffhall, line 61: Revised to make a little clearer.

_Holder of new auth token and/or owner(s) of new addresses for admin, clerk, operator must opt-in to the update._

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -98,73 +100,212 @@ contract SellerHandlerFacet is SellerBase {
...
-                (check1 == 0 || check1 == _seller.id) &&
-                    (check2 == 0 || check2 == _seller.id) &&
-                    (check3 == 0 || check3 == _seller.id),
-                SELLER_ADDRESS_MUST_BE_UNIQUE
+        if (_authToken.tokenType != AuthTokenType.None) {
+            // If AuthToken data is different from the one in storage, then set it as pending update
+            if (authToken.tokenType != _authToken.tokenType || authToken.tokenId != _authToken.tokenId) {
+                // Check that auth token is unique to this seller
+                uint256 check = lookups.sellerIdByAuthToken[_authToken.tokenType][_authToken.tokenId];
+                require(check == 0 || check == _seller.id, AUTH_TOKEN_MUST_BE_UNIQUE);
```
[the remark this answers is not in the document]
anajuliabit: I don't think we should remove it from `preUpdateSellerCheck`. What if a seller requests to update an operator address, but then this same address was used to create or update another seller id before optInToSellerUpdate is called for the first seller?

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -98,73 +100,212 @@ contract SellerHandlerFacet is SellerBase {
...
+        (, Seller storage seller, AuthToken storage authToken) = fetchSeller(_sellerId);
+
+        for (uint256 i = 0; i < _fieldsToUpdate.length; i++) {
+            SellerFields role = _fieldsToUpdate[i];
+
+            // Approve admin update
+            if (role == SellerFields.Admin && sellerPendingUpdate.admin != address(0)) {
+                require(sellerPendingUpdate.admin == sender, UNAUTHORIZED_CALLER_UPDATE);
+
+                preUpdateSellerCheck(_sellerId, sellerPendingUpdate.admin, lookups);
```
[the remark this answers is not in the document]
anajuliabit: done

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -98,73 +100,212 @@ contract SellerHandlerFacet is SellerBase {
...
+                // Delete old seller id by admin mapping
+                delete lookups.sellerIdByAdmin[seller.admin];
+
+                // Update admin
+                seller.admin = sellerPendingUpdate.admin;
```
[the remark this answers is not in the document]
anajuliabit: done

contracts/protocol/libs/ProtocolLib.sol
```diff
@@ -174,6 +174,10 @@ library ProtocolLib {
         mapping(uint256 => BosonTypes.Condition) exchangeCondition;
         // groupId => offerId => index on Group.offerIds array
         mapping(uint256 => mapping(uint256 => uint256)) offerIdIndexByGroup;
+        // seller id => Seller
+        mapping(uint256 => BosonTypes.Seller) sellerPendingUpdates;
+        // seller id => AuthToken
+        mapping(uint256 => BosonTypes.AuthToken) sellerPendingAuthTokenUpdate;
```
cliffhall, line 180: I'd like to suggest alternatives for these mappings that are more akin to the "X by Y" naming convention.

- `pendingAddressUpdatesBySeller`
- `pendingAuthTokenUpdatesBySeller`

scripts/domain/SellerFields.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * Boson Protocol Domain Enum: SellerFields
+ */
+class SellerFields {}
```
cliffhall, line 4: Perhaps this can be named `SellerUpdateFields`, since it is very specific to seller updates.

cliffhall (changes requested):
Just a few requests

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -56,7 +56,9 @@ contract SellerHandlerFacet is SellerBase {
      *         All other fields should be filled, even those staying the same.
```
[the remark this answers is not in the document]
anajuliabit: fixed

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -98,73 +100,212 @@ contract SellerHandlerFacet is SellerBase {
...
+
+    /**
+     * @notice Opt-in to a pending seller update
+     *
+     * Emits a SellerUpdateApplied event if successful.
+     *
+     * Reverts if:
+     * - The sellers region of protocol is paused
+     * - Addresses are not unique to this seller
+     * - Caller is not the address pending update for the field being updated
```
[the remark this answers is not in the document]
anajuliabit: done

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -56,7 +56,9 @@ contract SellerHandlerFacet is SellerBase {
      *         All other fields should be filled, even those staying the same.
      * @dev    Active flag passed in by caller will be ignored. The value from storage will be used.
      *
-     * Emits a SellerUpdated event if successful.
+     * Emits a SellerUpdateApplied event if the seller has changed the treasury.
+     * Emits a SellerUpdatePending event if the seller has requested an update for admin, clerk, operator, or auth token.
+     * Addresses owner of new values for admin, clerk, operator or auth token need to opt-in to update.
      *
      * Reverts if:
      * - The sellers region of protocol is paused
```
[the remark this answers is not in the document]
anajuliabit: What do you mean by "Caller is not the supplied admin"? This doesn't mean the same as "Caller is not the admin address of the stored seller"?

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -98,73 +100,212 @@ contract SellerHandlerFacet is SellerBase {
...
-        delete lookups.sellerIdByAuthToken[authToken.tokenType][authToken.tokenId];
-        delete protocolEntities().authTokens[seller.id];
+        // Get seller pending update
+        (
+            bool exists,
+            Seller storage sellerPendingUpdate,
+            AuthToken storage authTokenPendingUpdate
+        ) = fetchSellerPendingUpdate(_sellerId);
+
+        require(exists, NO_SELLER_PENDING_UPDATE);
```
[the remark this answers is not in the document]
anajuliabit: There is no way to exist a pending record if the seller doesn't exist. This check would be necessary if we had a way to delete seller accounts

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -98,73 +100,212 @@ contract SellerHandlerFacet is SellerBase {
...
+
+                // Delete pending update auth token
+                delete authTokenPendingUpdate.tokenType;
+                delete authTokenPendingUpdate.tokenId;
+            }
         }
 
         // Notify watchers of state change
-        emit SellerUpdated(_seller.id, _seller, _authToken, sender);
+        emit SellerUpdateApplied(_sellerId, seller, sellerPendingUpdate, authToken, authTokenPendingUpdate, sender);
```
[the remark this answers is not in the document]
anajuliabit: this is intentional. We want to show that there is nothing pending anymore emitting the events with zero values for `sellerPendingUpdate` and `authTokenPendingUpdate`

contracts/protocol/facets/SellerHandlerFacet.sol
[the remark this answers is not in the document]
anajuliabit: Remediate. I have also removed from `preUpdateSellerCheck` and moved the calls to within the if statements 
`if(seller.role !== _seller.role)`

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -98,73 +100,212 @@ contract SellerHandlerFacet is SellerBase {
...
+                require(tokenIdOwner == sender, UNAUTHORIZED_CALLER_UPDATE);
 
-        // Ignore active flag passed in by caller and set to value in storage.
-        _seller.active = seller.active;
-        storeSeller(_seller, _authToken);
+                // Check that auth token is unique to this seller
+                uint256 check = lookups.sellerIdByAuthToken[authTokenPendingUpdate.tokenType][
+                    authTokenPendingUpdate.tokenId
+                ];
+                require(check == 0 || check == _sellerId, AUTH_TOKEN_MUST_BE_UNIQUE);
```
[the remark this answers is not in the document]
anajuliabit: Remediate

contracts/protocol/facets/SellerHandlerFacet.sol
anajuliabit: done

contracts/protocol/libs/ProtocolLib.sol
anajuliabit: renamed

contracts/interfaces/handlers/IBosonAccountHandler.sol
```diff
@@ -97,17 +97,20 @@ interface IBosonAccountHandler is IBosonAccountEvents {
     function createAgent(BosonTypes.Agent memory _agent) external;
 
     /**
-     * @notice Updates a seller, with the exception of the active flag.
-     *         All other fields should be filled, even those staying the same.
+     * @notice Updates treasury address, if changed. Puts admin, operator, clerk and AuthToken in pending state, if changed.
```
cliffhall, line 100: Since the protocol has several formal state machines, I would use the word “queue” rather than “state” to avoid any confusion.

contracts/protocol/bases/ProtocolBase.sol
```diff
@@ -308,6 +308,40 @@ abstract contract ProtocolBase is PausableBase, ReentrancyGuardBase {
...
 
+    /**
+     * @notice Fetches a given seller pending update from storage by id
+     *
+     * @param _sellerId - the id of the seller
+     * @return exists - whether the seller or auth token pending update exists
+     * @return sellerPendingUpdate - the seller pending update details. See {BosonTypes.Seller}
+     * @return authTokenPendingUpdate - auth token pending update details
+     */
+    function fetchSellerPendingUpdate(uint256 _sellerId)
```
cliffhall, line 319: Does this really need to live in ProtocolBase? Could it instead go in SellerBase? Seems a bit specific to this use case to need to be inherited by every facet.

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -52,17 +53,20 @@ contract SellerHandlerFacet is SellerBase {
     }
 
     /**
-     * @notice Updates a seller, with the exception of the active flag.
-     *         All other fields should be filled, even those staying the same.
+     * @notice Updates treasury address, if changed. Puts admin, operator, clerk and AuthToken in pending state, if changed.
```
cliffhall, line 56: Since the protocol has several formal state machines, I would use the word “queue” rather than “state” to avoid any confusion.

cliffhall (changes requested):

contracts/interfaces/handlers/IBosonAccountHandler.sol
anajuliabit: renamed

contracts/protocol/facets/SellerHandlerFacet.sol
anajuliabit: renamed

contracts/protocol/bases/ProtocolBase.sol
anajuliabit: moved

contracts/protocol/facets/SellerHandlerFacet.sol
hswopeams: This one is fixed, but we still have a different order on when checking admin on line 136 and treasury on 161. Let's be consistent. This is the kind of thing the auditor will comment on as a "style" issue.

contracts/protocol/facets/SellerHandlerFacet.sol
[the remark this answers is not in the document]
hswopeams: You're right. The first one doesn't make sense. It should also revert if "Caller is not the address of the owner of the stored AuthToken". We need to make it clear that the updateSeller reverts if it is not called by the admin/owner of AuthToken of the current stored data.

contracts/protocol/facets/SellerHandlerFacet.sol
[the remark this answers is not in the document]
hswopeams: Let's discuss in the standup (and maybe during Engineering Congress) what information the SellerUpdateApplied should convey. I was thinking it should show that what was pending was actually stored for the seller. I think we should discuss as a team and with the core components team to see what info they need/want.

zajck (dismissed):

contracts/protocol/facets/SellerHandlerFacet.sol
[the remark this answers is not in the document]
hswopeams: Agreed during Engineering Congress that the current approach is the right one.

contracts/protocol/facets/SellerHandlerFacet.sol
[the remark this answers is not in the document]
anajuliabit: done

contracts/protocol/facets/SellerHandlerFacet.sol
anajuliabit: done

hswopeams (approved):

zajck (approved):

contracts/interfaces/handlers/IBosonAccountHandler.sol
```diff
@@ -97,17 +97,20 @@ interface IBosonAccountHandler is IBosonAccountEvents {
...
+     * Emits a SellerUpdatePending event if the seller has requested an update for admin, clerk, operator, or auth token.
+     * Holder of new auth token and/or owner(s) of new addresses for admin, clerk, operator must opt-in to the update.
      *
      * Reverts if:
      * - The sellers region of protocol is paused
      * - Address values are zero address
      * - Addresses are not unique to this seller
-     * - Caller is not the admin address of the seller
+     * - Caller is not the admin address of the stored seller
+     * - Caller is not the address of the owner of the stored AuthToken
```
cliffhall, line 113: I think these revert reasons can be made clearer

- Caller address is not the admin address of the stored seller with no AuthToken
- Caller is not the of the owner of the seller's stored AuthToken

contracts/interfaces/handlers/IBosonAccountHandler.sol
```diff
@@ -118,6 +121,24 @@ interface IBosonAccountHandler is IBosonAccountEvents {
...
+    /**
+     * @notice Opt-in to a pending seller update
+     *
+     * Emits a SellerUpdateApplied event if successful.
+     *
+     * Reverts if:
+     * - The sellers region of protocol is paused
+     * - Addresses are not unique to this seller
+     * - Caller is not the address pending update for the field being updated
+     * - Caller is not the address of the owner of the pending AuthToken being updated
```
cliffhall, line 133: I think these revert reasons can be made clearer

- Caller address is not pending for the field being updated
- Caller is not the of the owner of the pending AuthToken being updated

hswopeams (dismissed):

cliffhall (changes requested):
Just some suggested rewording of revert reasons

zajck (approved):

hswopeams (dismissed):

contracts/interfaces/handlers/IBosonAccountHandler.sol
```diff
@@ -97,17 +97,20 @@ interface IBosonAccountHandler is IBosonAccountEvents {
...
+     * Emits a SellerUpdatePending event if the seller has requested an update for admin, clerk, operator, or auth token.
+     * Holder of new auth token and/or owner(s) of new addresses for admin, clerk, operator must opt-in to the update.
      *
      * Reverts if:
      * - The sellers region of protocol is paused
      * - Address values are zero address
      * - Addresses are not unique to this seller
-     * - Caller is not the admin address of the seller
+     * - Caller address is not the admin address of the stored seller with no AuthToken
+     * - Caller is not the address of the owner of the seller's stored AuthToken
```
cliffhall, line 113: Caller is not the owner of the seller's stored AuthToken

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -52,17 +52,20 @@ contract SellerHandlerFacet is SellerBase {
...
+     * Emits a SellerUpdatePending event if the seller has requested an update for admin, clerk, operator, or auth token.
+     * Holder of new auth token and/or owner(s) of new addresses for admin, clerk, operator must opt-in to the update.
      *
      * Reverts if:
      * - The sellers region of protocol is paused
      * - Address values are zero address
      * - Addresses are not unique to this seller
-     * - Caller is not the admin address of the seller
+     * - Caller address is not the admin address of the stored seller with no AuthToken
+     * - Caller is not the address of the owner of the seller's stored AuthToken
```
cliffhall, line 68: Caller is not the owner of the seller's stored AuthToken

cliffhall (changes requested):
One more natspec change

hswopeams (approved):

zajck (approved):

cliffhall (approved):
LGTM! 👍

PR #430, opened 2022-10-08
pre-release-v2.0.0-rc.3

anajuliabit:
Deploy to mumbai


hswopeams (changes requested):
On second thought, I think we need to change the ABI directory to abis/mumbai/test/interfaces

so we can also have abis/mumbai/staging/interfaces

I think we should also change the address file to 80001-mumbai-test.json. Adding "test" or "staging" to the file name could be a manual step.

zajck (approved):
LGTM

hswopeams (approved):

PR #431, opened 2022-10-10
Pre release v2.0.0 rc.3 Mumbai Staging Deployment

hswopeams:
These are the details of the rc.3 Mumbai stating deployment. 
After agreement from @mischat and @zajck, I stopped the process in the middle of the verification step. All of the contracts had already been verified since we deployed the same bytecode Mumai already for the test env. The verification step had been running for more than an hour and was only on the 5th out of 24 contracts. I added a note to the log file.


zajck (approved):
LGTM

PR #433, opened 2022-10-10
Reverts when auth token is Custom on create and update seller

anajuliabit:
This PR reverts the transaction when the value passed on `authToken.tokenType` is Custom on create and update seller

### DIFF
--- a/contracts/protocol/facets/SellerHandlerFacet.sol
+++ b/contracts/protocol/facets/SellerHandlerFacet.sol
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 pragma solidity 0.8.9;
 
-import "hardhat/console.sol";
 import "../../domain/BosonConstants.sol";
 import { IBosonVoucher } from "../../interfaces/clients/IBosonVoucher.sol";
 import { SellerBase } from "../bases/SellerBase.sol";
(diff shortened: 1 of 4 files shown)


cliffhall (approved):
LGTM! 👍

cliffhall (changes requested):
~~Should also do this check and revert in `createDisputeResolver` and `updateDisputeResolver`~~

cliffhall (approved):
LGTM! 👍

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -84,6 +86,8 @@ contract SellerHandlerFacet is SellerBase {
             ADMIN_OR_AUTH_TOKEN
         );
```
mischat, line 88: Should we not ensure that the check happens when the seller.admin == address(0) and when the AuthTokenType is not AuthTokenType.None ?

mischat: Ok, apologies this does make sense to me. 

At this point in the code the address is either address(0) and NOT AuthTokenType.None OR is it NOT address(0) and AuthTokenType.None... 

which implies that if at this line, if it is NOT Null, it will be rejected if .Custom ...

mischat (approved):

PR #434, opened 2022-10-10
This makes the util constants file more legible when it comes to dates

mischat:
... and times ... No semantic change, just makes the code easier to read.

### DIFF
--- a/test/util/constants.js
+++ b/test/util/constants.js
@@ -1,8 +1,8 @@
 // Some periods in milliseconds
-const oneDay = 86400; //  1 day in seconds
+const oneDay = 86400; // 1 day in seconds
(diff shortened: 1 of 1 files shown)


test/util/constants.js
```diff
@@ -1,8 +1,8 @@
 // Some periods in milliseconds
```
cliffhall, line 1: Should say seconds instead of milliseconds

cliffhall (changes requested):
Comment needs updating

test/util/constants.js
mischat: yes, pushing now

hswopeams (approved):

cliffhall (approved):
LGTM! 👍

PR #435, opened 2022-10-10
Use tip to ensure each transaction's inclusion in the next block. 

cliffhall:
## What is this
Refactor the deployment script to use tip rather than gas price and/or limit to ensure a each transaction's inclusion in the block.

## Why 
* For details of the strategy, see https://docs.alchemy.com/docs/maxpriorityfeepergas-vs-maxfeepergas

* Suffice it to say, supplying only an attractive tip amount is the best way to ensure inclusion in the next block.

## What changed
* In env.example
  - set CONFIRMATIONS to 3 (we cannot run with only one confirmation on polygon with high uncle rate. Probably this needs to be between 5 and 10, we'll see

* In deploy-protocol-*.js,
  - replaced gasLimit with maxPriorityFeePerGas

* In deploy-suite.js,
  - fetch tipMultiplier from environment
  - set tipSuggestion to amount that is always returned from ethers
  - create maxPriorityFeePerGas as bignumber with tipMultiplier times tipSuggestion
  - remove fetch of gasLimit from environment, no longer needed (pre-1559)
  - in all deployment calls that passed gasLimit, now pass maxPriorityFeePerGas

* In environments.js,
  - fetch tipMultiplier from process.env.TIP_MULTIPLIER

### DIFF
--- a/scripts/deploy-suite.js
+++ b/scripts/deploy-suite.js
@@ -16,7 +16,7 @@ const { deployProtocolClients } = require("./util/deploy-protocol-clients.js");
 const { deployProtocolConfigFacet } = require("./util/deploy-protocol-config-facet.js");
 const { deployProtocolHandlerFacets } = require("./util/deploy-protocol-handler-facets.js");
 const { verifyOnBlockExplorer, verifyOnTestEnv } = require("./util/report-verify-deployments");
-const { delay, deploymentComplete, writeContracts } = require("./util/utils");
+const { delay, deploymentComplete, getFees, writeContracts } = require("./util/utils");
 const AuthTokenType = require("../scripts/domain/AuthTokenType");
 
 /**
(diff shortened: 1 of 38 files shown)


zajck (dismissed):
In `deploy-suite.js` we have few more calls to the contracts (`grantRole`, `setVoucherBeaconAddress`, `setBeaconProxyAddress`, `setAuthTokenContract`, `renounceRole`), which should then also use `maxPriorityFeePerGas`

scripts/deploy-suite.js
```diff
@@ -2,8 +2,10 @@ const environments = require("../environments");
 const hre = require("hardhat");
 const ethers = hre.ethers;
 const network = hre.network.name;
-let gasLimit;
 const confirmations = environments.confirmations;
```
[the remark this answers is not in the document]
zajck: I already made this change.

scripts/util/deploy-protocol-client-proxies.js
```diff
@@ -19,18 +20,18 @@ const confirmations = environments.confirmations;
...
   // Destructure the protocol client implementations
   [bosonClientBeacon] = protocolClients;
 
   // Deploy the ClientProxy for BosonVoucher
   const ClientProxy = await ethers.getContractFactory("BeaconClientProxy");
-  const clientProxy = await ClientProxy.deploy({ gasLimit });
+  const clientProxy = await ClientProxy.deploy(await getFees(maxPriorityFeePerGas));
   await clientProxy.deployTransaction.wait(confirmations);
 
   // init instead of constructors
```
hswopeams, line 37: Do need this for the initialize call just below this on line 38?

cliffhall (commented):
LGTM! 👍

hswopeams (commented):
Just one question. See comment.

scripts/util/deploy-protocol-client-proxies.js
zajck: Added.

hswopeams (approved):

dennisfurrer (approved):

PR #441, opened 2022-10-17
Upgrade script

zajck:
Script to add/upgrade/remove facets. It accepts the list of facets. If facet does not exist yet, it just adds all methods. If facets exists, it compares the method and adds new, replace existing and remove old that are not part of new facet. If facet is to be removed, it removes all selectors.

- config is a js file `scripts/config/facet-upgrade.js` (similar as for manage-roles)
- I didn't include autmatic contract verification, which is in line with the current `deploy-suite`.
- Script updates addresses and make a separate update log.
- I added npm scripts and updated the documentation.


hswopeams (changes requested):

scripts/upgrade-facets.js
```diff
@@ -0,0 +1,244 @@
...
+    switch (answer.toLowerCase()) {
+      case "y":
+      case "yes":
+        break;
+      case "n":
+      case "no":
+        process.exit(1);
+        break;
+      default:
+        break;
```
cliffhall, line 59: I would make the default do `process.exit()` as well. If they didn't give a yes or no answer, then we should just quit.

scripts/upgrade-facets.js
```diff
@@ -0,0 +1,244 @@
...
+    let selectorsToRemove = registeredSelectors.filter((value) => !selectorsToReplace.includes(value)); // unique old selectors
+    let selectorsToAdd = newSelectors.filter((value) => !selectorsToReplace.includes(value)); // unique new selectors
+
+    // Skip selectors if set in config
+    const selectorsToSkip = Facets.skip[newFacet.name] ? Facets.skip[newFacet.name] : [];
+    selectorsToReplace = removeSelectors(selectorsToReplace, selectorsToSkip);
+    selectorsToRemove = removeSelectors(selectorsToRemove, selectorsToSkip);
+    selectorsToAdd = removeSelectors(selectorsToAdd, selectorsToSkip);
+
+    // Logs
```
cliffhall, line 142: I would move this logging in this section to after the actual work is done, otherwise if that fails, then this output might be confusing.

cliffhall (changes requested):
Just a few notes

scripts/upgrade-facets.js
```diff
@@ -0,0 +1,244 @@
...
+        process.exit(1);
+        break;
+      default:
+        break;
+    }
+  }
+
+  // Get the accounts
+  const accounts = await ethers.provider.listAccounts();
+  const admin = accounts[0];
```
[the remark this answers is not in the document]
zajck: Yes, good point. I changed it that it's read from `.env.`

I wouldn't say that calling is `manage-roles.js` is really a part of the process. It's more like prerequisite that admin has the upgrader role. I'll add that to docs.

scripts/upgrade-facets.js
```diff
@@ -0,0 +1,244 @@
...
+  const hasRole = await accessController.hasRole(Role.UPGRADER, admin);
+  if (!hasRole) {
+    console.log("Admin address does not have UPGRADER role");
+    process.exit(1);
+  }
+
+  // Deploy new facets
+  const deployedFacets = await deployProtocolHandlerFacets(
+    protocolAddress,
+    Facets.addOrUpgrade,
```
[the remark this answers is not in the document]
zajck: In both cases we need to deploy whole new facets, since the new code must somehow come to blockchain. So even if we are only upgrading a facet, a deployment must take place at some point. And I think it sensible to do it here and not have a separate procedure if it's only upgraded.

scripts/upgrade-facets.js
```diff
@@ -0,0 +1,244 @@
...
+    // Determine actions to be made
+    let selectorsToReplace = registeredSelectors.filter((value) => newSelectors.includes(value)); // intersection of old and new selectors
+    let selectorsToRemove = registeredSelectors.filter((value) => !selectorsToReplace.includes(value)); // unique old selectors
+    let selectorsToAdd = newSelectors.filter((value) => !selectorsToReplace.includes(value)); // unique new selectors
+
+    // Skip selectors if set in config
+    const selectorsToSkip = Facets.skip[newFacet.name] ? Facets.skip[newFacet.name] : [];
+    selectorsToReplace = removeSelectors(selectorsToReplace, selectorsToSkip);
+    selectorsToRemove = removeSelectors(selectorsToRemove, selectorsToSkip);
+    selectorsToAdd = removeSelectors(selectorsToAdd, selectorsToSkip);
```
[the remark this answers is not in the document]
zajck: Yep, it doesn't really make sense to put something in the code and then skip it.
I removed this line.

scripts/upgrade-facets.js
zajck: Moved.

hswopeams (changes requested):
Still one open issue

docs/tasks.md
```diff
@@ -41,6 +41,20 @@ Deploy suite deploys protocol diamond, all facets, client and beacon, and initia
 - **Ethereum Mainnet**. This deploys the built contracts to Ethereum Mainnet.
 ```npm run deploy-suite:ethereum:mainnet```
 
+### Upgrade facet
+Upgrade existing facets, add new facets or remove existing facets. We provide different npm scripts for different use cases. Script for Hardhat network does not exist since contracts are discrded after the deployment, so they cannot be upgraded.
```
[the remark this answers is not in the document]
zajck: fixed

docs/local-development.md
```diff
@@ -92,6 +92,19 @@ To get the examples how to use the admin to perform actions, refer to unit test
...
+To test the upgrade functionality, you first need to setup upgrader account as described in previous section.
+
+To perform the upgrade you then
+- Update some of the existing facets or create new one.
+- Update config file `scripts/config/facet-upgrade.js`:
+  - "addOrUpgrade" is the list of facets that will be upgraded or added,
+  - "remove": list of facets that will be completely removed
+  - "skip" allows you to specify methods that will be ignored during the process.
+- Update `version` in `package.json`. If the version in `package.json` matches the existing version in addresses file, you will have to explicitly confirm that you want to proceed.
+- Run `npm run upgrade-facets:local`. This will deploy new facets and make all necessary diamond cuts. It also updates the existing addresses file `addresses/31337-localhost.json` and outputs the upgrade log to the console.
```
cliffhall, line 105: maybe replace example address filename with template. `addresses/<chain-id>-<environment>.json`

scripts/util/deploy-protocol-handler-facets.js
```diff
@@ -31,25 +32,29 @@ async function deployProtocolHandlerFacets(diamond, facetNames, maxPriorityFeePe
...
-    const facetCut = getFacetAddCut(deployedFacet.contract, [initFunction]);
-    const transactionResponse = await diamondCutFacet.diamondCut(
-      [facetCut],
-      deployedFacet.contract.address,
-      callData,
-      await getFees(maxPriorityFeePerGas)
-    );
-    await transactionResponse.wait(confirmations);
+    // Cut all the facets into the diamond
+
```
cliffhall, line 45: remove empty line

scripts/upgrade-facets.js
```diff
@@ -0,0 +1,332 @@
...
+ *
+ * Prerequisite:
+ * - Admin must have UPGRADER role. Use `manage-roles.js` to grant it.
+ *
+ * Process:
+ *  1.  Edit scripts/config/facet-upgrade.js.
+ *  1a. Provide a list of facets that needs to be upgraded (field "addOrUpgrade") or removed completely (field "remove")
+ *  1b. Optionally you can specify which selectors should be ignored (field "skip"). You don't have to specify "initialize()" since it's ignored by default
+ *  2. Update protocol version in package.json. If not, script will prompt you to confirm that version remains unchanged.
+ *  2. Run the appropriate npm script in package.json to upgrde facets for a given network
```
anajuliabit, line 33:
```suggestion
 *  2. Run the appropriate npm script in package.json to upgrade facets for a given network
```

scripts/upgrade-facets.js
```diff
@@ -0,0 +1,332 @@
...
+
+  // If admin address is unspecified, exit the process
+  if (adminAddress == ethers.constants.AddressZero || !adminAddress) {
+    console.log("Admin address must not be zero address");
+    process.exit(1);
+  }
+  // Get signer for admin address
+  const adminSigner = await ethers.getSigner(adminAddress);
+
+  console.log("🔱 Admin account: ", adminAddress ? adminAddress : "not found" && process.exit());
```
anajuliabit, line 77: not a big deal, but the condition falsy on this line will never be achieved because of lines 71-73.

cliffhall (changes requested):
A few minor notes

anajuliabit (changes requested):
Looks very good

Just small requests

docs/local-development.md
zajck: I replaced that with template, but I still kept the example in since this is the expected filename if they are just following all the steps.

scripts/upgrade-facets.js
zajck: Fixed.

scripts/util/deploy-protocol-handler-facets.js
zajck: removed

anajuliabit (approved):
LGTM

hswopeams (approved):

cliffhall (approved):
LGTM! 👍

PR #442, opened 2022-10-17
Add opt-in feature when updating dispute resolver admin, clerk, operator 

anajuliabit:
This PR adds an opt-in feature when updating DR admin, clerk, operator to prevent front-running.

New update dispute resolver steps (except for treasury, metadaUri and escalationResponsePeriod, which is updated automatically):
1.DR requests to update a field (admin, operator, clerk)
2.Event DisputeResolverUpdatePending is emitted
3. Addresses owner's calls optInToDisputeResolverUpdate passing the disputeResolverId and the fields in which they are approving the update, e.g: 
```javascript
optInToDisputeResolverUpdate("1", [DisputeResolverUpdateFields.Admin, DisputeResolverUpdateFields.Operator])
```
4. Event DisputeResolverUpdateApplied emitted


contracts/protocol/facets/DisputeResolverHandlerFacet.sol
```diff
@@ -188,40 +193,184 @@ contract DisputeResolverHandlerFacet is IBosonAccountEvents, ProtocolBase {
...
+     * - Caller address is not pending update for the field being updated
+     * - No pending update exists for this dispute resolver
+     *
+     * @param _disputeResolverId - disputeResolver id
+     * @param _fieldsToUpdate - fields to update, see DisputeResolverUpdateFields enum
+     */
+    function optInToDisputeResolverUpdate(
+        uint256 _disputeResolverId,
+        DisputeResolverUpdateFields[] calldata _fieldsToUpdate
+    ) external disputeResolversNotPaused nonReentrant {
```
cliffhall, line 296: We aren't making any calls to external, untrusted contracts here. Do we need `nonReentrant`?

cliffhall (changes requested):
Just one note about the reentrancy guard.

contracts/protocol/facets/DisputeResolverHandlerFacet.sol
```diff
@@ -188,40 +193,184 @@ contract DisputeResolverHandlerFacet is IBosonAccountEvents, ProtocolBase {
...
+                lookups.disputeResolverIdByClerk[sender] = _disputeResolverId;
+
+                // Delete pending update clerk
+                delete disputeResolverPendingUpdate.clerk;
+            }
+        }
 
         // Notify watchers of state change
-        emit DisputeResolverUpdated(_disputeResolver.id, _disputeResolver, sender);
+        emit DisputeResolverUpdateApplied(_disputeResolverId, disputeResolver, disputeResolverPendingUpdate, sender);
```
zajck, line 373: One observation - it can happend that actually nothing is updated, but we still emit this event.
Yes, we do check in line 306 that some pending update exists. But that is true is some some fields is not zero.

Still it can happen that let's say `sellerPendingUpdate.admin` is not zero but operator and treasury are. So check in 306 will pass. However, I might put in ` _fieldsToUpdate` only `SellerUpdateFields.Operator` which is not supposed to be updated. But tx will be confirmed normally, so whoever called it they might think that operator update was succeful (if they don't verify the emitted log).

I think we should track that some update really took place and emit the even only if it did and return otherwise.

And I think we have the same situation in `optInToSellerUpdate`

zajck (changes requested):

contracts/protocol/facets/DisputeResolverHandlerFacet.sol
cliffhall: Ok. Given that a scenario exists, however ridiculous, I agree we should leave it in. 

This is an interesting potential reentrancy -- not reentering the same method, but calling back in to a different method in order to leverage the current msgSender() -- that almost makes every method potentially needing `nonReentrant` on it.

contracts/interfaces/handlers/IBosonAccountHandler.sol
```diff
@@ -159,13 +159,19 @@ interface IBosonAccountHandler is IBosonAccountEvents {
...
+     *
+     *         Update doesn't include DisputeResolverFees, allowed seller list or active flag.
      *         All DisputeResolver fields should be filled, even those staying the same.
      *         Use removeFeesFromDisputeResolver and addFeesToDisputeResolver to add and remove fees.
      *         Use addSellersToAllowList and removeSellersFromAllowList to add and remove allowed sellers.
+     *
      * @dev    Active flag passed in by caller will be ignored. The value from storage will be used.
      *
      * Emits a DisputeResolverUpdated event if successful.
+     * Emits a DisputeResolverUpdatePending event if the seller has requested an update for admin, clerk, operator, or auth token.
```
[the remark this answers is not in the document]
anajuliabit: done

contracts/protocol/facets/DisputeResolverHandlerFacet.sol
```diff
@@ -139,13 +138,19 @@ contract DisputeResolverHandlerFacet is IBosonAccountEvents, ProtocolBase {
...
+     *
+     *         Update doesn't include DisputeResolverFees, allowed seller list or active flag.
      *         All DisputeResolver fields should be filled, even those staying the same.
      *         Use removeFeesFromDisputeResolver and addFeesToDisputeResolver to add and remove fees.
      *         Use addSellersToAllowList and removeSellersFromAllowList to add and remove allowed sellers.
+     *
      * @dev    Active flag passed in by caller will be ignored. The value from storage will be used.
      *
      * Emits a DisputeResolverUpdated event if successful.
+     * Emits a DisputeResolverUpdatePending event if the seller has requested an update for admin, clerk, operator, or auth token.
```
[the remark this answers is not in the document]
anajuliabit: done

contracts/protocol/facets/DisputeResolverHandlerFacet.sol
```diff
@@ -188,40 +193,184 @@ contract DisputeResolverHandlerFacet is IBosonAccountEvents, ProtocolBase {
...
+            _disputeResolver.id
         );
 
-        // Delete current mappings
-        delete lookups.disputeResolverIdByOperator[disputeResolver.operator];
-        delete lookups.disputeResolverIdByAdmin[disputeResolver.admin];
-        delete lookups.disputeResolverIdByClerk[disputeResolver.clerk];
+        if (_disputeResolver.admin != disputeResolver.admin) {
+            preUpdateDisputeResolverCheck(_disputeResolver.id, _disputeResolver.admin, lookups);
+            require(_disputeResolver.admin != address(0), INVALID_ADDRESS);
```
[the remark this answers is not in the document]
anajuliabit: removed

contracts/protocol/facets/DisputeResolverHandlerFacet.sol
```diff
@@ -188,40 +193,184 @@ contract DisputeResolverHandlerFacet is IBosonAccountEvents, ProtocolBase {
...
+
+            // If clerk address exists, clerk address owner must approve the update to prevent front-running
+            disputeResolverPendingUpdate.clerk = _disputeResolver.clerk;
+            needsApproval = true;
+        }
+
+        if (needsApproval) {
+            // Notify watchers of state change
+            emit DisputeResolverUpdatePending(_disputeResolver.id, disputeResolverPendingUpdate, sender);
+        }
```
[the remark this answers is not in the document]
anajuliabit: remediated

contracts/protocol/facets/DisputeResolverHandlerFacet.sol
anajuliabit: done!

zajck (approved):
LGTM

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -209,110 +205,134 @@ contract SellerHandlerFacet is SellerBase {
 
         require(exists, NO_PENDING_UPDATE_FOR_ACCOUNT);
 
-        // Get storage location for seller
-        (, Seller storage seller, AuthToken storage authToken) = fetchSeller(_sellerId);
+        {
```
cliffhall, line 208: Maybe add a comment that this block solves for a stack to deep error.

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -209,110 +205,134 @@ contract SellerHandlerFacet is SellerBase {
```
cliffhall, line 205: Add 
`// Be sure an update is pending`

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -209,110 +205,134 @@ contract SellerHandlerFacet is SellerBase {
...
-        for (uint256 i = 0; i < _fieldsToUpdate.length; i++) {
-            SellerUpdateFields role = _fieldsToUpdate[i];
+            // Get storage location for seller
+            (, Seller storage seller, AuthToken storage authToken) = fetchSeller(_sellerId);
 
-            // Approve admin update
-            if (role == SellerUpdateFields.Admin && sellerPendingUpdate.admin != address(0)) {
-                require(sellerPendingUpdate.admin == sender, UNAUTHORIZED_CALLER_UPDATE);
+            // Cache protocol lookups and sender for reference
+            ProtocolLib.ProtocolLookups storage lookups = protocolLookups();
```
cliffhall, line 215: Since msgSender() is called a lot in this method, and each time it accesses storage (in the case of a meta tx), we should fetch it into a memory location here and reuse throughout. I realize you may be dealing with stack to deep here, so that may not be possible, but it would be good if we could do it.

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -209,110 +205,134 @@ contract SellerHandlerFacet is SellerBase {
...
-                // Delete pending update operator
-                delete sellerPendingUpdate.operator;
-            }
+                    // Update operator
+                    seller.operator = msgSender();
 
-            // Aprove clerk update
-            if (role == SellerUpdateFields.Clerk && sellerPendingUpdate.clerk != address(0)) {
-                require(sellerPendingUpdate.clerk == sender, UNAUTHORIZED_CALLER_UPDATE);
+                    // Transfer ownership of NFT voucher to new operator
```
cliffhall, line 255: `// Transfer ownership of NFT voucher contract to new operator`

test/protocol/SellerHandlerTest.js
```diff
@@ -20,7 +20,7 @@ const { mockSeller, mockAuthToken, mockVoucherInitValues, accountId } = require(
 /**
  *  Test the Boson Seller Handler
  */
-describe("SellerHandler", function () {
+describe.only("SellerHandler", function () {
```
cliffhall, line 23: Remove `.only`

cliffhall (changes requested):
Just a few minor notes

contracts/protocol/facets/SellerHandlerFacet.sol
anajuliabit: I had first created a local variable, but then I need to remove to fix stack to deep error

contracts/protocol/facets/SellerHandlerFacet.sol
anajuliabit: done

contracts/protocol/facets/SellerHandlerFacet.sol
anajuliabit: done

contracts/protocol/facets/SellerHandlerFacet.sol
anajuliabit: actually there is no need for this block, I fixed the stack to deep error removing local `sender` variable.

cliffhall (dismissed):
LGTM! 👍

contracts/protocol/facets/SellerHandlerFacet.sol
zajck: There is a way to go around the stack too deep error - not by wrapping the bulk of the code in {}, but rather to do it already at the begining, since `exists` is used only there and it's later not needed.

If you start the `optInToSellerUpdate` with
```
        // Get seller pending update
        Seller storage sellerPendingUpdate;
        AuthToken storage authTokenPendingUpdate;
        {
            bool exists;
            (exists, sellerPendingUpdate, authTokenPendingUpdate) = fetchSellerPendingUpdate(_sellerId);

            // Be sure an update is pending
            require(exists, NO_PENDING_UPDATE_FOR_ACCOUNT);
        }
        
        bool updateApplied;
        address sender = msgSender();
```
And then replace all `msgSender()` with `sender` that will work.

As a side note, even if that was not possible, then better than getting rid of `address seller` would be to sacrifice `SellerUpdateFields role`. Still better to read multiple times from the memory than from the storage.

contracts/protocol/facets/DisputeResolverHandlerFacet.sol
```diff
@@ -188,40 +193,194 @@ contract DisputeResolverHandlerFacet is IBosonAccountEvents, ProtocolBase {
...
+                lookups.disputeResolverIdByAdmin[sender] = _disputeResolverId;
+
+                // Delete pending update admin
+                delete disputeResolverPendingUpdate.admin;
+
+                updateApplied = true;
+            }
+
+            // Approve operator update
+            if (role == DisputeResolverUpdateFields.Operator && disputeResolverPendingUpdate.operator != address(0)) {
```
zajck, line 333: This should be `else if` since in each iteration `role` will match only one of possible fields.

Same is true in `optInToSellerUpdate`

zajck (changes requested):
I just noticed something else.

Also have a look at [my comment about stack too deep](https://github.com/bosonprotocol/boson-protocol-contracts/pull/442#discussion_r1001425544)

mischat:
just to confirm, that the the events that are emitted at the end of this new update flow are the same as the events that used to be emitted in the "1-step"  updates.

anajuliabit:
@mischat we decided to replace the `DisputeResolverUpdated` for `DisputeResolverUpdateApplied` (which is emitted wherever a DR field is changed)  and add a new event called `DisputeResolverUpdatePending`

contracts/protocol/facets/DisputeResolverHandlerFacet.sol
anajuliabit: nice catch, done

contracts/protocol/facets/SellerHandlerFacet.sol
anajuliabit: that makes sense!

zajck (dismissed):
LGTM :+1:

contracts/interfaces/handlers/IBosonAccountHandler.sol
```diff
@@ -159,13 +159,19 @@ interface IBosonAccountHandler is IBosonAccountEvents {
...
      *         Use addSellersToAllowList and removeSellersFromAllowList to add and remove allowed sellers.
+     *
      * @dev    Active flag passed in by caller will be ignored. The value from storage will be used.
      *
      * Emits a DisputeResolverUpdated event if successful.
+     * Emits a DisputeResolverUpdatePending event if the dispute resolver has requested an update for admin, clerk or operator.
+     * Owner(s) of new addresses for admin, clerk, operator must opt-in to the update.
      *
      * Reverts if:
      * - The dispute resolvers region of protocol is paused
```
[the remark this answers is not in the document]
hswopeams: Also, the "EscalationResponsePeriod is invalid" revert reason is missing here and in the handler

hswopeams (changes requested):
Here are some comments that I couldn't put inline because they are not shown in the update view:

"should update state correctly if values are the same" test case name on line 948 in DisputeResolverHandlerTest.js is no longer valid. We are now checking if values have changed and not updating if they have not. This test case should now be called something like 
"state should stay the same if values are the same". It looks like there is also a redundant  expect(returnedDisputeResolver.isValid()).is.true; check on line 961

Since updateDisputeResolver no longer calls storeDisputeResolve, the INVALID_ESCALATION_PERIOD check can be moved from storeDisputeResolver to createDisputeResolver

contracts/protocol/facets/SellerHandlerFacet.sol
cliffhall: Then why is this code still in a block? I checked git blame on the block opening curly and it was from a commit that said stack to deep error. Is there a reason to keep this in a block still?

cliffhall (changes requested):
Just a note about that weird code block.

contracts/interfaces/handlers/IBosonAccountHandler.sol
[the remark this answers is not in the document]
anajuliabit: done

contracts/interfaces/handlers/IBosonAccountHandler.sol
```diff
@@ -178,6 +184,25 @@ interface IBosonAccountHandler is IBosonAccountEvents {
...
 
+    /**
+     * @notice Opt-in to a pending dispute resolver update
+     *
+     * Emits a DisputeResolverUpdateApplied event if successful.
+     *
+     * Reverts if:
+     * - The dispute resolver region of protocol is paused
+     * - Addresses are not unique to this dispute resolver
+     * - Caller address is not pending update for the field being updated
```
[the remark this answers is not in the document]
anajuliabit: done

anajuliabit:
@cliffhall I changed the way we handle the stack to deep error on this method, based on @zajck suggestion https://github.com/bosonprotocol/boson-protocol-contracts/pull/442#discussion_r1001425544

contracts/protocol/facets/DisputeResolverHandlerFacet.sol
```diff
@@ -103,6 +103,14 @@ contract DisputeResolverHandlerFacet is IBosonAccountEvents, ProtocolBase {
             _disputeResolverFees.length <= protocolLimits().maxFeesPerDisputeResolver,
             INVALID_AMOUNT_DISPUTE_RESOLVER_FEES
         );
+
+        // Escalation period must be greater than zero and less than or equal to the max allowed
+        require(
+            _disputeResolver.escalationResponsePeriod > 0 &&
+                _disputeResolver.escalationResponsePeriod <= protocolLimits().maxEscalationResponsePeriod,
```
cliffhall, line 110: If it doesn't cause a stack too deep error, we should probably fetch the protocolLimits() storage slot to a local var and reuse it for these two requires.

cliffhall (changes requested):
Just a few notes

hswopeams (dismissed):

contracts/interfaces/handlers/IBosonAccountHandler.sol
```diff
@@ -148,7 +148,7 @@ interface IBosonAccountHandler is IBosonAccountEvents {
      *
      * Reverts if:
      * - The buyers region of protocol is paused
-     * - Caller is not the wallet address associated with the buyer account
+     * - Caller is not the admin address of the stored buyer
```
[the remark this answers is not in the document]
hswopeams: Agreed. Should be "wallet". This was a change due to a request I made to make the reference to stored buyer consistent with seller and DR update funcitons.

contracts/protocol/facets/DisputeResolverHandlerFacet.sol
anajuliabit: done

zajck (approved):
LGTM :+1:

contracts/protocol/facets/DisputeResolverHandlerFacet.sol
```diff
@@ -96,13 +93,28 @@ contract DisputeResolverHandlerFacet is IBosonAccountEvents, ProtocolBase {
             DISPUTE_RESOLVER_ADDRESS_MUST_BE_UNIQUE
         );
 
-        _disputeResolver.id = disputeResolverId;
+        {
```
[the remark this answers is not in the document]
cliffhall: I also see the same pattern at line 63.

contracts/protocol/facets/DisputeResolverHandlerFacet.sol
```diff
@@ -188,40 +207,192 @@ contract DisputeResolverHandlerFacet is IBosonAccountEvents, ProtocolBase {
         // Check that msg.sender is the admin address for this dispute resolver
```
cliffhall, line 207: "Check that caller is the admin..."

cliffhall (changes requested):
Unexplained unconditional code blocks.

contracts/protocol/facets/DisputeResolverHandlerFacet.sol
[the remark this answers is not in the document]
anajuliabit: I added the comments

contracts/protocol/facets/DisputeResolverHandlerFacet.sol
anajuliabit: done

cliffhall: It doesn't look like this change actually got made.

cliffhall (changes requested):
One skipped item

cliffhall (approved):
LGTM! 👍

zajck (approved):

PR #446, opened 2022-10-21
Update GitHub actions

anajuliabit:
This PR upgrade some GitHub actions to make Node 16 compatible and also replace deprecated command `set-output`


.github/workflows/ci.yaml
```diff
@@ -16,13 +16,13 @@ jobs:
       test-chunks:  ${{ steps['set-test-chunks'].outputs['test-chunks'] }}
       test-chunk-ids: ${{ steps['set-test-chunk-ids'].outputs['test-chunk-ids'] }}
     steps:
-      - uses: webfactory/ssh-agent@v0.5.4
+      - uses: webfactory/ssh-agent@v0.7.0
         with:
           ssh-private-key: ${{ secrets.SSH_GITHUB_SOLIDOC }}
```
zajck, line 21: I'm pretty sure we don't need this anymore, since solidoc repo is now public. I suggest you remove it.
Probably whole ssh-agent can be removed everywhere, since all our dependencies are public now.

zajck (commented):
I see only `ci.yaml` was changed. Did you deploy-pre-release.yaml check also `deploy-pre-release.yaml` and versions there are ok?

.github/workflows/ci.yaml
```diff
@@ -16,13 +16,13 @@ jobs:
...
       - name: Checkout repository
-        uses: actions/checkout@v2.4.0
+        uses: actions/checkout@v3
```
hswopeams, line 23: Does @v3 get the latest version?

hswopeams:
I see new deprecation errors. now. We'll have to see if these changes fix this. I'm seeing the same errors on the faucet now too.

[build](https://github.com/bosonprotocol/boson-protocol-contracts/actions/runs/3312682463/jobs/5469656583#step:11:8)
The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

.github/workflows/ci.yaml
anajuliabit: removed

.github/workflows/ci.yaml
anajuliabit: I could not find any minor version on the docs; every example I have seen uses only `v3`. It's weird because the repo release says that the latest is 2.5.0 https://github.com/actions/checkout, not sure if I understood how the actions versions works

anajuliabit:
@hswopeams I don't know from where those warning was coming, but it seems to be fixed after my last commit

zajck (approved):
LGTM

.github/workflows/ci.yaml
hswopeams: The way the releases show up is a bit weird. I don't know why it shows v2.5.0 as the latest. If you go here, you see that 3.1.0 is actually the latest: https://github.com/actions/checkout/releases. Maybe 2.5.0 is just the latest for v2.

anajuliabit: I changed to the latest both checkup and setup-node actions. I had tried with 3.1.x or 3.x.x but it didn't work

hswopeams (approved):

PR #447, opened 2022-10-25
Deploy or upgrade facets with initialize arguments

zajck:
So far, deployment and upgrade script assumed that all facets have initalizers without arguments. The only exception was `configHandlerFacet` which has its own script `deploy-protocol-config-facet`.

This PR introduces new method `deployProtocolHandlerFacetsWithArgs` where you can pass in the initializer arguments.
Existing methods `deployProtocolHandlerFacets` and `deployProtocolConfigFacet` were refactored so now they both use `deployProtocolHandlerFacetsWithArgs`. They were still kept in, since they are used all over the code and replacing them everywhere would be tedious and error prone.

I had to slightly modify `scripts/config/facet-upgrade.js` and I added a new config file `scripts/config/facet-deploy.js`

In this branch only the configHandler accepts the arguments, so you can test only with it. In my other branch https://github.com/bosonprotocol/boson-protocol-contracts/tree/whitelist-metatx-initializer `MetaTransactionHandler` also accepts init args, and that branch has the same deploy/upgrade script as this, so you can test there as well.

### DIFF
--- a/docs/local-development.md
+++ b/docs/local-development.md
@@ -100,7 +100,9 @@ To perform the upgrade you then
 - Update config file `scripts/config/facet-upgrade.js`:
   - "addOrUpgrade" is the list of facets that will be upgraded or added,
   - "remove": list of facets that will be completely removed
-  - "skip" allows you to specify methods that will be ignored during the process.
+  - "skipSelectors" allows you to specify methods that will be ignored during the process.
+  - "initArgs": if facet initializer expects arguments, provide it here. For no-arg initializers you don't have to specify anything.
+  - "skipInit": list of facets for which you want to skip initialization call.
 - Update `version` in `package.json`. If the version in `package.json` matches the existing version in addresses file, you will have to explicitly confirm that you want to proceed.
 - Run `npm run upgrade-facets:local`. This will deploy new facets and make all necessary diamond cuts. It also updates the existing addresses file `addresses/<chain-id>-<environment>.json` (for example `addresses/31337-localhost.json` if you are using a default local hardhat node) and outputs the upgrade log to the console.
 
--- a/docs/local-development.md
+++ b/docs/local-development.md
@@ -101,7 +101,7 @@ To perform the upgrade you then
   - "addOrUpgrade" is the list of facets that will be upgraded or added,
   - "remove": list of facets that will be completely removed
   - "skipSelectors" allows you to specify methods that will be ignored during the process.
-  - "initArgs": if facet initializer expects arguments, provide it here. For no-arg initializers you don't have to specify anything.
+  - "initArgs": if facet initializer expects arguments, provide them here. For no-arg initializers you don't have to specify anything.
   - "skipInit": list of facets for which you want to skip initialization call.
 - Update `version` in `package.json`. If the version in `package.json` matches the existing version in addresses file, you will have to explicitly confirm that you want to proceed.
 - Run `npm run upgrade-facets:local`. This will deploy new facets and make all necessary diamond cuts. It also updates the existing addresses file `addresses/<chain-id>-<environment>.json` (for example `addresses/31337-localhost.json` if you are using a default local hardhat node) and outputs the upgrade log to the console.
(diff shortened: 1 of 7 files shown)


anajuliabit (dismissed):
I ran the scripts in your other branch and on this one as well. Everything works fine.

LGTM

docs/local-development.md
```diff
@@ -100,7 +100,9 @@ To perform the upgrade you then
 - Update config file `scripts/config/facet-upgrade.js`:
   - "addOrUpgrade" is the list of facets that will be upgraded or added,
   - "remove": list of facets that will be completely removed
-  - "skip" allows you to specify methods that will be ignored during the process.
+  - "skipSelectors" allows you to specify methods that will be ignored during the process.
+  - "initArgs": if facet initializer expects arguments, provide it here. For no-arg initializers you don't have to specify anything.
```
hswopeams, line 104:
```suggestion
  - "initArgs": if facet initializer expects arguments, provide them here. For no-arg initializers you don't have to specify anything.
```

scripts/config/facet-deploy.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * Config file used to upgrade the facets
```
hswopeams, line 2:
```suggestion
 * Config file used to deploy the facets
```

scripts/config/facet-upgrade.js
```diff
@@ -3,12 +3,16 @@
  *
  * - addOrUpgrade: list of facets that will be upgraded or added
  * - remove: list of facets that will be completely removed
- * - skip:  mapping "facetName":"listOfFunctionsToBeSkipped". With this you can specify functions that will be ignored during the update.
+ * - skipSelectors:  mapping "facetName":"listOfFunctionsToBeSkipped". With this you can specify functions that will be ignored during the update.
  *          You don't have to specify "initialize()" since it's ignored by default.
  *          Skip does not apply to facets that are completely removed.
+ * - initArgs: if facet initializer expects arguments, provide it here. For no-arg initializers you don't have to specify anything.
```
hswopeams, line 9:
```suggestion
 * - initArgs: if facet initializer expects arguments, provide them here. For no-arg initializers you don't have to specify anything.
```

scripts/config/facet-upgrade.js
```diff
@@ -3,12 +3,16 @@
...
+ * - skipInit": list of facets for which you want to skip initialization call.
  */
 exports.Facets = {
-  addOrUpgrade: ["SellerHandlerFacet", "BuyerHandlerFacet"],
+  addOrUpgrade: [],
```
hswopeams, line 13: I think it makes sense to remove the examples from the actual code, but I think it's a good idea to have an example config in the comments. This is for all of the options.

scripts/config/facet-deploy.js
```diff
@@ -0,0 +1,26 @@
...
+    "ExchangeHandlerFacet",
+    "FundsHandlerFacet",
+    "GroupHandlerFacet",
+    "OfferHandlerFacet",
+    "OrchestrationHandlerFacet",
+    "TwinHandlerFacet",
+    "PauseHandlerFacet",
+    "MetaTransactionsHandlerFacet",
+  ],
+  argFacets: {},
```
hswopeams, line 25: I think it's a good idea to have an example config in the comments.

hswopeams (changes requested):
Looks good. I have a few requests regarding the comments.

docs/local-development.md
zajck: Done.

scripts/config/facet-deploy.js
zajck: Done.

scripts/config/facet-deploy.js
zajck: Done.

scripts/config/facet-upgrade.js
zajck: Done

scripts/config/facet-upgrade.js
zajck: Done.

hswopeams (approved):

anajuliabit (approved):

PR #451, opened 2022-10-28
Remove gas limit

anajuliabit:


hswopeams (changes requested):
hardhat.config.js still refers to `environments.hardhat.gasLimit`. The reference should be removed EXCPET for the hardhat network. If this value is not set (which points to `process.env.DEPLOYER_GAS_LIMIT_TEST` in environments.js), the following unit test error occurs:
```
IBosonOfferHandler
       📋 Offer Handler Methods - BATCH
         👉 createOfferBatch()
           💔 Revert Reasons
             Creating too many offers:
     AssertionError: Expected transaction to be reverted with Exceeded maximum offers in a single transaction, but other exception was thrown: InvalidInputError: Transaction gas limit is 30069144 and exceeds block gas limit of 30000000
```
setting DEPLOYER_GAS_LIMIT_TEST=8000000 fixes the error.

It appears that the `deployMockTokens` and `deployAndMintMockNFTAuthTokens` functions is also still using gasLimit. The `deployMockTokens` function is only called from test files. However, the test files pass in an invalid value. The test files get the value using` const { gasLimit } = require("../../environments");`, which is undefined. So `deployMockTokens` isn't really using gasLimit. The correct way to get the gasLimit would be

```
const environments = require("../../environments");
const gasLimit = environments[network].gasLimit;
```
As can be seen in the `deployAndMintMockNFTAuthTokens` function.

I think the best course of action is to keep the gas property in hardhat.conig.js (and add it back to environments.js and .env) and remove from other network configs. In addition, the gasLimit can be removed from the `deployMockTokens` and `deployAndMintMockNFTAuthTokens` functions. Since the tests are passing "undefined" as gasLimit to `deployMockTokens` , it can be removed from the tests as well

zajck:
> hardhat.config.js still refers to `environments.hardhat.gasLimit` in the network configs. The reference should be removed EXCEPT for the hardhat network. If this value is not set (which points to `process.env.DEPLOYER_GAS_LIMIT_TEST` in environments.js), the following unit test error occurs:
> 
> ```
> IBosonOfferHandler
>        📋 Offer Handler Methods - BATCH
>          👉 createOfferBatch()
>            💔 Revert Reasons
>              Creating too many offers:
>      AssertionError: Expected transaction to be reverted with Exceeded maximum offers in a single transaction, but other exception was thrown: InvalidInputError: Transaction gas limit is 30069144 and exceeds block gas limit of 30000000
> ```
> 
> setting DEPLOYER_GAS_LIMIT_TEST=8000000 in the .env file fixes the error.

This is the only test case with the error, right?
I suggest that we remove `environments.hardhat.gasLimit` for hardhat network, too, and set a manual gas limit only in this particular test.

hswopeams:
> > hardhat.config.js still refers to `environments.hardhat.gasLimit` in the network configs. The reference should be removed EXCEPT for the hardhat network. If this value is not set (which points to `process.env.DEPLOYER_GAS_LIMIT_TEST` in environments.js), the following unit test error occurs:
> > ```
> > IBosonOfferHandler
> >        📋 Offer Handler Methods - BATCH
> >          👉 createOfferBatch()
> >            💔 Revert Reasons
> >              Creating too many offers:
> >      AssertionError: Expected transaction to be reverted with Exceeded maximum offers in a single transaction, but other exception was thrown: InvalidInputError: Transaction gas limit is 30069144 and exceeds block gas limit of 30000000
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > setting DEPLOYER_GAS_LIMIT_TEST=8000000 in the .env file fixes the error.
> 
> This is the only test case with the error, right? I suggest that we remove `environments.hardhat.gasLimit` for hardhat network, too, and set a manual gas limit only in this particular test.
For me it's only this one.

cliffhall (dismissed):
LGTM! 👍

hswopeams (changes requested):
It looks mostly OK, except that the hardhat config file still has a gas property for all networks. We had discussed only keeping it for the hardhat network. @zajck's suggestion might be better though.

cliffhall (approved):
LGTM! 👍

cliffhall (approved):
LGTM! 👍

hswopeams (approved):

PR #454, opened 2022-11-02
Code coverage

anajuliabit:
**This PR adds new unit tests to cover uncovered branches and statements and remove some unachieved code on contracts**

The impossible test cases were described here https://docs.google.com/document/d/1K4YGbpV4QevqxfsJQOfsJAisco2udJ-ltC94MCQK8wk/edit

### Contract changes 

- Remove the if statement on
```solidity
      if (_bundle.offerIds.length > 0) {
                 bundleSupplyChecks(offersTotalQuantityAvailable, twinId);
       }
```

because `offerIds` length has to be greater than 0 (we revert above on line 51 when offerIds length is 0). See https://github.com/bosonprotocol/boson-protocol-contracts/blob/fbb731bfb513b5e281bf8aadb425a5831338a332/contracts/protocol/bases/BundleBase.sol#L51-L53

- Remove the if statement on 
```solidity
    if (_role != address(0)) {
            uint256 check1 = _lookups.disputeResolverIdByOperator[_role];
            uint256 check2 = _lookups.disputeResolverIdByClerk[_role];
            uint256 check3 = _lookups.disputeResolverIdByAdmin[_role];

            require(
                (check1 == 0 || check1 == _disputeResolverId) &&
                    (check2 == 0 || check2 == _disputeResolverId) &&
                    (check3 == 0 || check3 == _disputeResolverId),
                DISPUTE_RESOLVER_ADDRESS_MUST_BE_UNIQUE
            );
        }
```
because on `updateDisputeResolver` we're reverting when any address is 0
https://github.com/bosonprotocol/boson-protocol-contracts/blob/fbb731bfb513b5e281bf8aadb425a5831338a332/contracts/protocol/facets/DisputeResolverHandlerFacet.sol#L189-L195
and on `optInToUpdateDisputeResolver` we have an if statement that prevents from calling that function with role = address(0)
https://github.com/bosonprotocol/boson-protocol-contracts/blob/fbb731bfb513b5e281bf8aadb425a5831338a332/contracts/protocol/facets/DisputeResolverHandlerFacet.sol#L327

- Remove the revert on 
https://github.com/bosonprotocol/boson-protocol-contracts/blob/fbb731bfb513b5e281bf8aadb425a5831338a332/contracts/protocol/facets/SellerHandlerFacet.sol#L135
since when admin address is 0 means that auth token has been used and we never achieve this line 

## Notes

I don't know if its a bug in the code coverage library but yesterday the code coverage on BeaconClientProxy and Proxy contracts were reported with very low coverage so I was writing unit tests on `BeaconClientProxy.js` but today, when I ran it again they were reported with 100% coverage. 

[coverage-all.zip](https://github.com/bosonprotocol/boson-protocol-contracts/files/9938469/coverage-all.zip)

### DIFF
--- a/test/access/AccessControllerTest.js
+++ b/test/access/AccessControllerTest.js
@@ -304,7 +304,7 @@ describe("AccessController", function () {
     });
 
     it("Should not emit 'RoleRevoked' event if revoking a role that was not granted", async function () {
-      // Renounce Role, should not emit the event
+      // Revoke Role, should not emit the event
       await expect(accessController.connect(admin).revokeRole(Role.ADMIN, rando.address)).to.not.emit(
         accessController,
         "RoleRevoked"
@@ -452,7 +452,7 @@ describe("AccessController", function () {
     });
 
     it("Should revert if caller tries to grantRole but doesn't have ADMIN role", async function () {
-      // Renounce Role, expecting revert
+      // Grant Role, expecting revert
       await expect(accessController.connect(rando).grantRole(Role.ADMIN, rando.address)).to.be.revertedWith(
         `AccessControl: account ${rando.address.toLowerCase()} is missing role ${Role.ADMIN}`
       );
--- a/test/access/AccessControllerTest.js
+++ b/test/access/AccessControllerTest.js
@@ -304,7 +304,7 @@ describe("AccessController", function () {
     });
 
     it("Should not emit 'RoleRevoked' event if revoking a role that was not granted", async function () {
-      // Renounce Role, expecting the event
+      // Renounce Role, should not emit the event
       await expect(accessController.connect(admin).revokeRole(Role.ADMIN, rando.address)).to.not.emit(
         accessController,
         "RoleRevoked"
@@ -405,7 +405,7 @@ describe("AccessController", function () {
     });
 
     it("Should not emit 'RoleRevoked' event if renouncing a role that was not granted", async function () {
-      // Renounce Role, expecting the event
+      // Renounce Role, should not emit the event
       await expect(accessController.connect(rando).renounceRole(Role.ADMIN, rando.address)).to.not.emit(
         accessController,
         "RoleRevoked"
@@ -445,14 +445,14 @@ describe("AccessController", function () {
 
   context("💔 Revert Reasons", async function () {
     it("Caller is different from account to be renounced", async function () {
-      // Renounce Role, expecting the event
+      // Renounce Role, expecting revert
       await expect(accessController.connect(admin).renounceRole(Role.ADMIN, deployer.address)).to.be.revertedWith(
         RevertReasons.CAN_ONLY_REVOKE_SELF
       );
     });
 
-    it("Should revert if caller has no role", async function () {
-      // Renounce Role, expecting the event
+    it("Should revert if caller tries to grantRole but doesn't have ADMIN role", async function () {
+      // Renounce Role, expecting revert
       await expect(accessController.connect(rando).grantRole(Role.ADMIN, rando.address)).to.be.revertedWith(
         `AccessControl: account ${rando.address.toLowerCase()} is missing role ${Role.ADMIN}`
       );
--- a/test/access/AccessControllerTest.js
+++ b/test/access/AccessControllerTest.js
@@ -2,23 +2,43 @@ const hre = require("hardhat");
 const ethers = hre.ethers;
 const { expect } = require("chai");
 const Role = require("../../scripts/domain/Role");
+const { getInterfaceIds } = require("../../scripts/config/supported-interfaces.js");
+const { RevertReasons } = require("../../scripts/config/revert-reasons.js");
 
 /**
  *  Test the AccessController contract
  */
 describe("AccessController", function () {
   // Shared args
-  let deployer, admin, protocol, upgrader, associate, pauser, client, feeCollector;
+  let deployer, admin, protocol, upgrader, associate, pauser, client, feeCollector, rando;
   let AccessController, accessController, roleAdmin;
+  let InterfaceIds;
+
+  before(async function () {
+    // get interface Ids
+    InterfaceIds = await getInterfaceIds();
+  });
 
   beforeEach(async function () {
     // Make accounts available
-    [deployer, admin, protocol, upgrader, associate, pauser, client, feeCollector] = await ethers.getSigners();
+    [deployer, admin, protocol, upgrader, associate, pauser, client, feeCollector, rando] = await ethers.getSigners();
 
     // Deploy the contract
     AccessController = await ethers.getContractFactory("AccessController");
     accessController = await AccessController.deploy();
     await accessController.deployed();
+  });
+
+  context("📋 Interfaces", async function () {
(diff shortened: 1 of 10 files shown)


test/access/AccessControllerTest.js
```diff
@@ -282,6 +302,17 @@ describe("AccessController", function () {
         "ADMIN role can't revoke FEE_COLLECTOR role"
       ).is.false;
     });
+
+    it("Should not emit 'RoleRevoked' event if revoking a role that was not granted", async function () {
+      // Renounce Role, should not emit the event
```
zajck, line 307:
```suggestion
      // Revoke Role, should not emit the event
```

test/access/AccessControllerTest.js
```diff
@@ -400,4 +442,20 @@ describe("AccessController", function () {
...
+  context("💔 Revert Reasons", async function () {
+    it("Caller is different from account to be renounced", async function () {
+      // Renounce Role, expecting revert
+      await expect(accessController.connect(admin).renounceRole(Role.ADMIN, deployer.address)).to.be.revertedWith(
+        RevertReasons.CAN_ONLY_REVOKE_SELF
+      );
+    });
+
+    it("Should revert if caller tries to grantRole but doesn't have ADMIN role", async function () {
+      // Renounce Role, expecting revert
```
zajck, line 455:
```suggestion
      // Grant Role, expecting revert
```

test/protocol/DisputeResolverHandlerTest.js
```diff
@@ -2454,6 +2454,27 @@ describe("DisputeResolverHandler", function () {
           );
       });
 
+      it("Should not emit 'DisputeResolverUpdateApplied' event if caller doesn't especify any field", async function () {
```
zajck, line 2457:
```suggestion
      it("Should not emit 'DisputeResolverUpdateApplied' event if caller doesn't specify any field", async function () {
```

test/protocol/DisputeResolverHandlerTest.js
```diff
@@ -2454,6 +2454,27 @@ describe("DisputeResolverHandler", function () {
...
+        disputeResolver.operator = other1.address;
+        await accountHandler.connect(admin).updateDisputeResolver(disputeResolver);
+
+        await expect(accountHandler.connect(other1).optInToDisputeResolverUpdate(disputeResolver.id, [])).to.not.emit(
+          accountHandler,
+          "DisputeResolverUpdateApplied"
+        );
+      });
+
+      it("Should not emit 'DisputeResolverUpdateApplied'event if there is no pending update for especified field", async function () {
```
zajck, line 2467:
```suggestion
      it("Should not emit 'DisputeResolverUpdateApplied' event if there is no pending update for specified field", async function () {
```

test/protocol/SellerHandlerTest.js
```diff
@@ -2577,6 +2631,25 @@ describe("SellerHandler", function () {
           );
       });
 
+      it("Should not emit 'SellerUpdateApplied' event if caller doesn't especify any field", async function () {
```
zajck, line 2634:
```suggestion
      it("Should not emit 'SellerUpdateApplied' event if caller doesn't specify any field", async function () {
```

test/protocol/SellerHandlerTest.js
```diff
@@ -2577,6 +2631,25 @@ describe("SellerHandler", function () {
...
+        seller.operator = other1.address;
+        await accountHandler.connect(admin).updateSeller(seller, emptyAuthToken);
+
+        await expect(accountHandler.connect(other1).optInToSellerUpdate(seller.id, [])).to.not.emit(
+          accountHandler,
+          "SellerUpdateApplied"
+        );
+      });
+
+      it("Should not emit 'SellerUpdateApplied'event if there is no pending update for especified field", async function () {
```
zajck, line 2644:
```suggestion
      it("Should not emit 'SellerUpdateApplied' event if there is no pending update for specified field", async function () {
```

zajck (changes requested):
Looks good, I just found some typos.

And please add `ext_libs/` to `skipFiles` in `.solcover.js`

test/protocol/SellerHandlerTest.js
```diff
@@ -2181,6 +2208,33 @@ describe("SellerHandler", function () {
           );
```
hswopeams, line 2208: Just above this on line 2200, there's a test case "addresses are the zero address". There's something not quite right with this test case. It looks like it's meant to check if the INVALID_ADDRESS revert gets triggered when all the addresses are zero. However, it doesn't set the treasury address to zero. It DOES set the admin address to zero, but it also passes in a valid authToken (not emptyAuthToken). So a zero admin address would be valid.  A different revert happens if the admin address/auth token combo is not correct (ADMIN_OR_AUTH_TOKEN). I think to make this a valid use case, the line 

`seller.admin = ethers.constants.AddressZero;` on line 2202 should be changed to 
`seller.treasury = ethers.constants.AddressZero;`

There doesn't seem to be a corresponding test case in the createSeller Revert Reasons context

hswopeams (changes requested):
Looks pretty good. I just found a few small things.

test/protocol/SellerHandlerTest.js
anajuliabit: fixed

test/protocol/SellerHandlerTest.js
anajuliabit: fixed

test/protocol/DisputeResolverHandlerTest.js
anajuliabit: fixed

test/protocol/DisputeResolverHandlerTest.js
anajuliabit: fixed

test/access/AccessControllerTest.js
anajuliabit: fixed

test/access/AccessControllerTest.js
anajuliabit: fixed

test/protocol/DisputeResolverHandlerTest.js
zajck: There is still a space missing between `'` and `event`.

Since you need to push another change anyway (to fix failing test), can you please also change this as well.

test/protocol/SellerHandlerTest.js
hswopeams: The problem in the test case starting on line 2200 has been fixed. There is still no corresponding "addresses are the zero address" test case for createSeller. To be consistent, it should go just above  the "Operator address is zero address" test case on line 909.

hswopeams (changes requested):

test/protocol/SellerHandlerTest.js
anajuliabit: sorry I overlooked your comments

hswopeams (approved):

zajck (approved):

PR #457, opened 2022-11-03
Add interfaceIds to addesses file

zajck:
Old files with addresses did not include interface ids. I added them for smoother upgrades.

### DIFF
--- a/addresses/80001-mumbai-staging.json
+++ b/addresses/80001-mumbai-staging.json
@@ -6,22 +6,26 @@
     {
       "name": "AccessController",
       "address": "0x955AbDE524CEd6A8AcA3A4a097bd3484c09f58bE",
-      "args": []
+      "args": [],
+      "interfaceId": ""
     },
     {
       "name": "DiamondLoupeFacet",
       "address": "0x120643A4c4EDcd2e86C67193006B1e0c44EEF1A9",
-      "args": []
+      "args": [],
+      "interfaceId": "0x48e2b093"
     },
     {
       "name": "DiamondCutFacet",
       "address": "0x77Ca7EBAAfC81433e46601e1acA7c245B33e2421",
-      "args": []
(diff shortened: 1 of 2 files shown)


addresses/80001-mumbai-staging.json
```diff
@@ -59,105 +63,125 @@
...
+      "interfaceId": ""
     },
     {
       "name": "BosonVoucher Proxy",
       "address": "0xC54A578dbcBD57d844976eABDA61B853d0a0a724",
-      "args": []
+      "args": [],
+      "interfaceId": ""
     }
   ]
```
hswopeams, line 186: Do we not have a way of getting the interface Ids of the voucher contracts, or do we not need them?

hswopeams (commented):
Did you get the interface Ids from the https://github.com/bosonprotocol/boson-protocol-contracts/releases/tag/v2.0.0 tag?

We should probably change the deployment script to write the interfaceId, but we could do that later.

addresses/80001-mumbai-staging.json
zajck: Technically we could get them, but we don't really need them, since they live outside the diamond and will never be part of diamond upgrade.

zajck:
> Did you get the interface Ids from the https://github.com/bosonprotocol/boson-protocol-contracts/releases/tag/v2.0.0 tag?
Yes, interface ids are from 2.0.0 deployment

> We should probably change the deployment script to write the interfaceId, but we could do that later.
It already does it. I added that at the same time when I created upgrade script. But since the 2.0.0 deployment were done before that, addresses file have the old format. If you look at addresses file for 2.1.0-rc.1 deployment, all interface ids are already included in the file.

hswopeams (approved):

anajuliabit (approved):

PR #461, opened 2022-11-07
Whitelist metatx methods

### LINKED ISSUE #460: Potential selector collision in generic metatransactions
### Description:

Metatransactions include among others two fields:
- `functionName`: human readable function name
- `functionSignature`: abi encoded data that is passed to protocol to be executed.

These two values are connected, such that first 4 bytes of `keccak256(abi.encodePacked(functionName))` match first 4 bytes of `functionSignature`. This relationship is validated here:

https://github.com/bosonprotocol/boson-protocol-contracts/blob/de03b0655ae10e0a21ed0949e17af0dd0a4e34c5/contracts/protocol/facets/MetaTransactionsHandlerFacet.sol#L191

However, since only 4 bytes determine the function selector, it is possible to find a collision, i.e. `functionName` that does not correspond to `functionSignature`, but still has the same selector. In this case user might sign a metatransaction thinking that they function in `functionName` will be executed, but truly some other function (set in non human readable `functionSignature`) will be executed.

### Recommendation:

Add allow list for functions that can be executed via meta transactions, where we can verify that all 32 bytes of `keccak256(abi.encodePacked(functionName))` belong to the function and not first 4.

zajck:
When metatransaction is validated, we check that function from argument `_functionName` is whitelisted and revert if it's not.

Code changes:
- modified `validateTx`
- added setter `setWhitelistedFunctions` and getter `isFunctionWhitelisted`
- white list is implemented as `mapping bytes32=>bool`, where key is `keccak256(_functionName)`. This is more efficient that storing `mapping string=>bool`. It is also in line how roles are represented in accessControl.

Fixes #460


contracts/protocol/facets/MetaTransactionsHandlerFacet.sol
```diff
@@ -305,4 +316,41 @@ contract MetaTransactionsHandlerFacet is IBosonMetaTransactionsHandler, Protocol
...
+    {
+        ProtocolLib.ProtocolMetaTxInfo storage pmti = protocolMetaTxInfo();
+
+        // set new values
+        for (uint256 i = 0; i < _functionNameHashes.length; i++) {
+            pmti.isWhitelisted[_functionNameHashes[i]] = _isWhitelisted;
+        }
+
+        // Notify external observers
+        emit FunctionsWhitelisted(_functionNameHashes, _isWhitelisted, msgSender());
```
hswopeams, line 344: I agree strings aren't efficient to store, but if we only emit the keccak256 hash of the function, we'll never really be able to see which functions are white listed. It might be better to pass in and emit an array of strings but  hash and store them in the function as bytes32 hash.

test/protocol/MetaTransactionsHandlerTest.js
```diff
@@ -637,8 +716,85 @@ describe("IBosonMetaTransactionsHandler", function () {
...
+                r,
+                s,
+                v
+              )
+            ).to.revertedWith(RevertReasons.FUNCTION_NOT_WHITELISTED);
+          });
+
           it("Should fail when function name is incorrect", async function () {
-            let incorrectFunctionName = "createSeller"; // there are no function argument types here.
+            let incorrectFunctionName = "redeemVoucher(uint256)"; // function name is whitelisted, but different than what we encode in next step
```
hswopeams, line 797: Is this test case sufficient to test the collision you pointed out in your document between redeemVoucher and cancelVoucher?

test/protocol/MetaTransactionsHandlerTest.js
```diff
@@ -939,7 +1103,7 @@ describe("IBosonMetaTransactionsHandler", function () {
               s,
               v
             )
-          ).to.revertedWith(RevertReasons.INVALID_FUNCTION_SIGNATURE);
+          ).to.revertedWith(RevertReasons.FUNCTION_NOT_WHITELISTED);
         });
 
         context("Reentrancy guard", async function () {
```
hswopeams, line 1109: General comment: `isFunctionWhitelisted` is an external function, so there should be some separate test cases for it.

hswopeams (changes requested):

test/protocol/MetaTransactionsHandlerTest.js
zajck: I added a test that test actual collision case

test/protocol/MetaTransactionsHandlerTest.js
zajck: I added an `isFunctionWhitelisted` context

scripts/config/facet-upgrade.js
```diff
@@ -21,10 +23,38 @@
...
+    ["executeMetaTransaction(address,string,bytes,uint256,bytes32,bytes32,uint8)"]
+  );
+
+  return {
+    addOrUpgrade: ["MetaTransactionsHandlerFacet"],
+    remove: [],
+    skipSelectors: { SellerHandlerFacet: [] },
+    initArgs: {
+      MetaTransactionsHandlerFacet: [MetaTransactionsHandlerFacetInitArgs],
+    },
```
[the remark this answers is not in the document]
hswopeams: `skipSelectors: { SellerHandlerFacet: [] }` is still in there. It was probably there before but should be removed as well.

contracts/protocol/facets/MetaTransactionsHandlerFacet.sol
hswopeams: Changing `isFunctionWhitelisted` to take a string could be a good solution.Otherwise, I'm afraid we're going to be completely blind. Let's discuss with the team.

hswopeams (changes requested):
Looks pretty good. There are just a few outstanding points.

hswopeams (approved):

contracts/interfaces/handlers/IBosonMetaTransactionsHandler.sol
```diff
@@ -54,29 +54,29 @@ interface IBosonMetaTransactionsHandler is IBosonMetaTransactionsEvents {
...
      *
-     * Emits a FunctionsWhitelisted event if successful.
+     * Emits a FunctionsAllowlisted event if successful.
      *
      * Reverts if:
      * - Caller is not a protocol admin
      *
      * @param _functionNameHashes - a list of hashed function names (keccak256)
-     * @param _isWhitelisted - new whitelist status
+     * @param _isAllowlisted - new allowl status
```
hswopeams, line 63:
```suggestion
     * @param _isAllowlisted - new allowlist status
```

hswopeams (changes requested):
One typo. Otherwise looks good.

hswopeams (dismissed):

test/protocol/MetaTransactionsHandlerTest.js
```diff
@@ -637,8 +834,85 @@ describe("IBosonMetaTransactionsHandler", function () {
             ).to.revertedWith(RevertReasons.REGION_PAUSED);
           });
 
+          it("Should fail when function name is not whitelised", async function () {
```
cliffhall, line 837: ~~whitelised~~ allowlisted

test/protocol/MetaTransactionsHandlerTest.js
```diff
@@ -637,8 +834,85 @@ describe("IBosonMetaTransactionsHandler", function () {
...
+                functionSignature,
+                nonce,
+                r,
+                s,
+                v
+              )
+            ).to.revertedWith(RevertReasons.FUNCTION_NOT_ALLOWLISTED);
+          });
+
+          it("Should fail when function name is not whitelised - incorrect name", async function () {
```
cliffhall, line 877: ~~whitelised~~ allowlisted

cliffhall (changes requested):
Just a few test descriptions

test/protocol/MetaTransactionsHandlerTest.js
zajck: fixed

test/protocol/MetaTransactionsHandlerTest.js
zajck: fixed

hswopeams (approved):

cliffhall (approved):
LGTM! 👍

PR #462, opened 2022-11-07
Specify environment in scripts

zajck:
So far scripts distinguished different network, but to use them in two different environments on a same network, manual manipulation of filenames was required.

This PR adds 3 hardhat tasks that accept `env` as an optional parameter:
- `deploy-suite`
- `manage-roles`
- `upgrade-facets`

Change was small since only write/read of addresses file needed to be modified.

I also updated npm tasks and task.md.


anajuliabit (approved):
LGTM

package.json
```diff
@@ -32,29 +32,32 @@
...
+    "deploy-suite:hardhat": "npx hardhat clean && npx hardhat compile && npx hardhat deploy-suite --network hardhat",
+    "deploy-suite:local": "npx hardhat clean && npx hardhat compile && npx hardhat deploy-suite --network localhost",
+    "deploy-suite:test": "npx hardhat clean && npx hardhat compile && npx hardhat deploy-suite --network test >> logs/test.deploy.contracts.txt",
+    "deploy-suite:ethereum:mainnet": "npx hardhat clean && npx hardhat compile && npx hardhat deploy-suite --network mainnet >> logs/mainnet.deploy.contracts.txt",
+    "deploy-suite:polygon:mumbai-test": "npx hardhat clean && npx hardhat compile && npx hardhat deploy-suite --network mumbai --env test >> logs/mumbai-test.deploy.contracts.txt",
+    "deploy-suite:polygon:mumbai-staging": "npx hardhat clean && npx hardhat compile && npx hardhat deploy-suite --network mumbai --env staging >> logs/mumbai-staging.deploy.contracts.txt",
+    "deploy-suite:polygon:mainnet": "npx hardhat clean && npx hardhat compile && npx hardhat deploy-suite --network polygon >> logs/polygon.deploy.contracts.txt",
```
hswopeams, line 41: Don't we need --env prod (or similar) for mainnet deployments? Will it work without "env" being passed in?
And --env test for deployment to the test env.

package.json
```diff
@@ -32,29 +32,32 @@
...
     "verify-suite:ethereum:mainnet": "npx hardhat verify-suite --network mainnet --chain-id 1 --env mainnet  >> logs/mainnet.verify.contracts.txt",
     "verify-suite:polygon:mumbai-test": "npx hardhat verify-suite --network mumbai --chain-id 80001 --env mumbai-test >> logs/mumbai-test.verify.contracts.txt",
```
hswopeams, line 43: I ran `npm run deploy-suite:polygon:mumbai-test`. The address file was created correctly. I then ran `npm run verify-suite:polygon:mumbai-test`. It finished very quickly without verifying anything. The only thing in the lig file was "🔍 Verifying contracts on block explorer..."
Usually even if the contract bytecode has already been verified, it takes a while and then it reports an error.

hswopeams (commented):
See comments inline

package.json
zajck: I don't think this has anything to do with the changes I made.
I checked that `scripts/util/verify-suite.js` still reads file as expected.

Verify script accepts filter which tells which contracts should be verified and the default filter is an empty array, so it just skips all verifications.

package.json
```diff
@@ -32,29 +32,32 @@
...
-    "manage-roles:polygon:mumbai": "npx hardhat --network mumbai run scripts/manage-roles.js >> logs/mumbai.manage.roles.txt",
-    "manage-roles:polygon:mainnet": "npx hardhat --network polygon run scripts/manage-roles.js >> logs/polygon.manage.roles.txt",
-    "manage-roles:local": "npx hardhat --network localhost run scripts/manage-roles.js >> logs/localhost.manage.roles.txt",
-    "manage-roles:test": "npx hardhat --network test run scripts/manage-roles.js >> logs/test.manage.roles.txt",
+    "upgrade-facets:local": "npx hardhat clean && npx hardhat compile && npx hardhat upgrade-facets --network localhost",
+    "upgrade-facets:test": "npx hardhat clean && npx hardhat compile && npx hardhat upgrade-facets --network test >> logs/test.upgrade.contracts.txt",
+    "upgrade-facets:ethereum:mainnet": "npx hardhat clean && npx hardhat compile && npx hardhat upgrade-facets --network mainnet >> logs/mainnet.upgrade.contracts.txt",
+    "upgrade-facets:polygon:mumbai-test": "npx hardhat clean && npx hardhat compile && npx hardhat upgrade-facets --network mumbai --env test >> logs/mumbai-test.upgrade.contracts.txt",
```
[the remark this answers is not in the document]
zajck: This won't be a problem for the staging upgrade. 
Script compares version in `package.json` (2.1.0) to the version in `80001-mumbai-staging.json` (2.0.0-rc.5) so it will pass without the prompt.

But still thanks for spotting this, since in some cases this could really be a problem.
For a moment I solved it by changing prompt type from stdout so stderr which is not redirected to the file and normally shows in the terminal.
Alternatively we could use `| tee` to always output everything to terminal and file at the same time.

scripts/util/utils.js
```diff
@@ -18,20 +18,20 @@ function deploymentComplete(name, address, args, interfaceId, contracts) {
...
-  const path = getAddressesFilePath(chainId, env);
+  const network = hre.network.name;
+  const path = getAddressesFilePath(chainId, network, env);
   fs.writeFileSync(
     path,
     JSON.stringify(
       {
         chainId: chainId,
-        env: env || "",
+        env: network || "",
```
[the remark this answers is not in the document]
zajck: I added it.
I don't think we actively use this field anywhere but it's nice to have all the data available.

PR #469, opened 2022-11-18
Fix bug on createTwinInternal

anajuliabit:
Remediate bug on createTwinInternal


contracts/protocol/bases/TwinBase.sol
```diff
@@ -138,7 +139,7 @@ contract TwinBase is ProtocolBase, IBosonTwinEvents {
         twin.tokenType = _twin.tokenType;
 
         // Notify watchers of state change
-        emit TwinCreated(twinId, sellerId, _twin, sender);
+        emit TwinCreated(twin.id, sellerId, _twin, sender);
```
zajck, line 142: This should stay `twinId` as it was to avoid reading from storage.

zajck (changes requested):
One small request.

contracts/protocol/bases/TwinBase.sol
anajuliabit: fixed

hswopeams (approved):

zajck (approved):

PR #468, opened 2022-11-17
Verify succesful upgrade - new actions after upgrade

zajck:
Continues work from #465 

It addresses three points that were not implemented there
- [x] Verfiy that adding new data after the upgrade does not corrupt the old data
- [x] Verify that actions that use data from before the upgrade still work as expected
- [x] Verify breaking changes, i.e. calls that should be broken after the upgrade really behave differently than before the update

PR #471, opened 2022-11-22
Verify succesful upgrade - refactor

zajck:
Refactor of code in #468 

- utility functions to deploy, upgrade, populate contract, get contract data are moved to `test/util/upgrade.js`
- reorganization of test variable - script no longer relies on positional arguments to distinguish between pre- and post- upgrade state
- deploy and upgrade configs are now in `test/upgrade/00_config.js` and is no longer required to modify default configs
- generic tests were extracted out of  `test/upgrade/2.0.0-2.1.0.js` and moved into a separate file `test/upgrade/01_generic.js` (it's named that way so it will always stay on top).

### DIFF
--- a/hardhat.config.js
+++ b/hardhat.config.js
@@ -45,18 +45,20 @@ task(
   "Deploy suite deploys protocol diamond, all facets, client and beacon, and initializes protcol diamond"
 )
   .addOptionalParam("env", "The deployment environment")
-  .setAction(async ({ env }) => {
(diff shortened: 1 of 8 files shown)


cliffhall (approved):
LGTM! 👍

cliffhall (approved):
LGTM! 👍

anajuliabit (approved):
Really nice!

PR #472, opened 2022-11-22
Update source of Solidoc, solc version and excluded files

Rhian:
This PR updates the contract documentation generation process. Included are:

- Update of `solc` in Truffle config to match pragma version in contracts
- Extended list of excluded contracts in `solidoc.json`
- Updated source for `solidoc` package in `package.json` now it has been published to the Boson npm registry


solidoc.json
```diff
@@ -3,36 +3,97 @@
...
     "Test2Facet.json",
+    "Test2FacetUpgrade.json",
     "Test3Facet.json",
+    "TestFacet256.json",
     "TestFacetLib.json",
-    "Test2FacetUpgrade.json" ]
+    "TestInitializableDiamond.json",
+    "Test2FacetUpgrade.json",
+    "TestProtocolFunctions.json" ]
 }
```
cliffhall, line 99: There is a new folder `contracts/example/` which has the snapshot gate example and support contracts. You may need to omit these as well.

cliffhall: Also, the ignorelist seems like it could get out of control. Is there no better way of specifying what to include?

cliffhall (approved):
Just some comments.

solidoc.json
Rhian: `solidoc` currently works on an exclusion basis... let me take a look to see if I can reverse it so that it works off an include-list

anajuliabit (approved):
LGTM

Can you fix the conflicts please @Rhian ?

solidoc.json
cliffhall: MIght be nice to have either mode available. Depending on whether the list to include would be longer that the one to exclude or visa versa.

Rhian: I've found a more sensible way to do this - we can specify the directories to compile in the `package.json` and that also allows us to then use the `solidoc.json` to exclude files within this. So currently the change to do this is in this repo rather than touching `solidoc`.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bosonprotocol/boson-protocol-contracts",
-  "version": "2.1.0",
+  "version": "2.0.0",
```
zajck, line 3: This version should stay "2.1.0"

zajck (changes requested):

cliffhall (approved):
Aside from fixing the version number in package-lock.json, LGTM! 👍

package.json
Rhian: Thanks @zajck - have updated this

zajck (approved):

cliffhall (approved):
LGTM! 👍

PR #473, opened 2022-11-23
Pre-minted voucher support

cliffhall:
## Status: WIP
* Ready to hand over the baton to @zajck as I'm out until Monday
* Modifies voucher contract to allow
  - reservation of a range 
    - not validated as to potential overlap with other offer ranges or existing token ids
      - will be called predictably by the diamond at pre-minted offer creation time
      - range will be relative to the next exchange id
      - not sure overlap is a concern
  - preminting of all or part of the reserved range for an offer
    - must be done by contract owner (seller/operator)
    - does not require caller to know the first id to mint
    - range must already be reserved
    - only emits the transfer events and stores the count that has been pre-minted for the range
  - getter to check how many remain in a range to be pre-minted
  - reporting of seller as owner of reserved token ids that aren't yet owned
  - getter to check if a token id is committable (reserved and preminted but not yet owned)
  - overrides of transferFrom and safeTransferFrom allow us to find out if the token is committible first, 
    - showstopper is we need to be able to set the seller as owner and mimic the supers, but the arrays they manage are private.
    
## Changes so far 
* In BeaconClientBase.sol,
  - added getBosonExchange method

* In BosonConstants.sol,
  - added revert reasons 
    - EXCHANGE_ID_IN_RESERVED_RANGE 
    - NO_RESERVED_RANGE_FOR_OFFER 
    - OFFER_RANGE_ALREADY_RESERVED 
    - INVALID_RANGE_START 
    - INVALID_AMOUNT_TO_MINT

* In BosonVoucher.sol,
  - added clarifications to contract-level doc about the use of the upgradeable contracts, which could be a red-herring for someone investigating this contract.
  - imported IBosonExchangeHandler
  - added Range struct
  - added rangeByOffer mapping
  - added ranges array
  - in issueVoucher,
    - added check to revert if exchange id falls within a reserved range
  - added methods 
    - reserveRange
      - only callable by PROTOCOL 
      - reverts if start id is zero 
      - reverts if offer id already has a range 
      - stores the range with a zero minted count
    - preMint 
      - only callable by contract owner (seller operator) 
      - takes offer id and amount to mint 
      - reverts if amount is greater than remaining amount to be pre-minted 
      - stores the new minted count for the offer 
    - getAvailablePreMints 
      - returns the number of vouchers left to be pre-minted for an offer 
    - getPreMintStatus 
      - checks if a token id is committable
      - Committable means
        - does not yet have an owner
        - in a reserved range
        - has been pre-minted
        - also returns associated offerId if committable
    - ownerOf 
      - override 
      - reports seller as owner if if token is reserved but not yet pre-minted 
      - reports actual owner if one exists 
      - reverts if not reserved and no owner is stored
  - overrode safeTransferFrom and transferFrom methods
    - check if committable and call super method if not
    - placeholder/todo if committable, 
      - we need to figure out how to set the seller as the owner **before** calling the super method, so that transfer will not revert because owner is address zero
  - in _beforeTokenTransfer method,
    - moved to bottom of contract with other internals
    - refactored to check for committable token and logic down to placeholder comment that calls commitToPreMintedOffer on protocol


zajck:
My changes only: https://github.com/bosonprotocol/boson-protocol-contracts/pull/473/files/ba396a7fbfe7299528160921c4439737631280cb..1d3e5dea7bbe4e51f80473da13154ff18cb7a341

## Changes so far 

* In BosonConstants.sol,
  - added revert reason 
    - NO_SILENT_MINT_ALLOWED

* In IBosonVoucher.sol
  - copied `Range` struct from BosonVoucher.sol
  - added methods to the interface:
    - reserveRange
    - preMint
    - getAvailablePreMints
    - getRangeByOfferId

* In BosonVoucher.sol,
  - moved `Range` struct to interface (maybe the best final location would be BosonTypes file)
  - defined struct `ERC721UpgradeableStorage` which helps when accessing private variables form ERC721Upgradeable
  - changed `Range[] private ranges` to `uint256[] private rangeOfferIds`. Before, both `rangeByOfferId` and `ranges` stored the full Range object. It's not only storage ineffcient, but when some range was updated (field `minted` during the pre mint), we'd need to update this value in both places to ensure consistency. Now `rangeOfferIds` only stores the offer ids, which allows to fetch the actual range when needed
  - added `premintStatus` variable that is used during the transfers and discarded afterwards
  - in reserveRange,
    - store offerd id to `rangeOfferIds`
    - emits event `RangeReserved`
  - in premint
    - updated some checks
    - update seller's balance (private variable `_balances`)
  - changed `ownerOf` to cover also the case of preminted and already burned voucher
  - in `transferFrom` and `safeTransferFrom`
    - calls `silentMint` which updates private variable `_owners`
    - store `premintStatus` which is then used by `_beforeTokenTransfer`
  - in `_beforeTokenTransfer`
    - it does not call `getPremintStatus` anymore, since `getPremintStatus` returns false if `_exists(tokenId)` is `true`. In `_beforeTokenTransfer` that's always the case since even if it's the first transfer of preminted token, `silentMint` changes token's ownership to seller. This has effect on `_exists()` which determines the existance based on ownership.
  - in getPreMintStatus
    - changed it to work with new variable `rangeByOfferId`
    - added a check, when token is found, that it wasn't burned already
  - added methods 
    - getRangeByOfferId
      - returns full range object
    - getERC721UpgradeableStorage
      - returns storage pointer to private variables in  `ERC721Upgradeable`
    - silentMint
      - changes ownership of a token to contract owner (seller) without emitting an event
      - it updates `_owners` variable
      - it reverts if called with `from` different than the owner
- TESTS:
  - in  revert-reasons.js
    - added new revert reasons
  -  added scripts and tests for `Range` entity
  - BosonVoucher.js
    - added tests for
      - reserveRange
      - preMint
      - getAvailablePreMints
      - getRange
      - ownerOf
      - all three token transfers methods

Still TODO:
- when `commitToPreMintedOffer` is implemented:
  - uncomment code in `_beforeTokenTransfer`
  - make proper test setup in "Transfer of a preminted voucher"
  - write currently skipped tests "Should call commitToPreMintedOffer" and "Second transfer should behave as normal voucher transfer"
- implement `maxPremintedTokens` (or something similar) to prevent running into a block gas limit in `preMint`. It's still to be discusedd where this value will be stored. Storing it in the voucher makes it's updating cumbersome. Maybe it would be sensible to store it in the protocol config together with other limits, and voucher contract gets it from there.

cliffhall (commented):
LGTM so far!

cliffhall (commented):
Looks good so far. Do you plan to add the protocol changes in the same PR or a new one?

zajck:
> Looks good so far. Do you plan to add the protocol changes in the same PR or a new one?

I'll make a separate PR for protocol changes.

contracts/protocol/clients/voucher/BosonVoucher.sol
```diff
@@ -77,6 +128,201 @@ contract BosonVoucher is IBosonVoucher, BeaconClientBase, OwnableUpgradeable, ER
...
+        // Revert if no more to mint in range
+        require(range.length >= range.minted + _amount, INVALID_AMOUNT_TO_MINT);
+
+        // Pre-mint the range to the seller
+        uint256 tokenId;
+        address seller = owner();
+        for (uint256 i = 0; i < _amount; i++) {
+            tokenId = start + i;
+            emit Transfer(address(0), seller, tokenId);
+        }
```
anajuliabit, line 217: You're not calling `_mint` function here because there is no mint at all, right? are you just reserving the tokenId to be used in the future? Just to make sure I understood

contracts/protocol/clients/voucher/BosonVoucher.sol
```diff
@@ -77,6 +128,201 @@ contract BosonVoucher is IBosonVoucher, BeaconClientBase, OwnableUpgradeable, ER
...
+
+        // Update seller's total balance
+        getERC721UpgradeableStorage()._balances[seller] += _amount;
+    }
+
+    /**
+     * @notice Gets the number of vouchers left to be pre-minted for an offer.
+     *
+     * @param _offerId - the id of the offer
+     * @return count - the count of pre-minted vouchers in reserved range
```
anajuliabit, line 230:
```suggestion
     * @return count - the count of available vouchers reserved in the range to be pre-minted 
```

contracts/protocol/clients/voucher/BosonVoucher.sol
```diff
@@ -77,6 +128,201 @@ contract BosonVoucher is IBosonVoucher, BeaconClientBase, OwnableUpgradeable, ER
...
+    function safeTransferFrom(
+        address from,
+        address to,
+        uint256 tokenId,
+        bytes memory data
+    ) public virtual override(ERC721Upgradeable, IERC721Upgradeable) {
+        (bool committable, uint256 offerId) = getPreMintStatus(tokenId);
+
+        if (committable) {
+            // If offer is committable, temporary update _owners, so transfer succedds
```
anajuliabit, line 317:
```suggestion
            // If offer is committable, temporary update _owners, so transfer succeed
```

contracts/protocol/clients/voucher/BosonVoucher.sol
```diff
@@ -279,4 +496,143 @@ contract BosonVoucher is IBosonVoucher, BeaconClientBase, OwnableUpgradeable, ER
...
+                        low = mid + 1;
+                    }
+                }
+            }
+        }
+    }
+
+    /*
+     * Returns storage pointer to location of private variables
+     * 0x0000000000000000000000000000000000000000000000000000000000000099 is location of _owners
```
anajuliabit, line 616: How `0x0000000000000000000000000000000000000000000000000000000000000099` could be both the location of _owners and ERC721UpgradeableStorage?

anajuliabit (changes requested):
Looks pretty good. 
Just small changes and some questions

contracts/protocol/clients/voucher/BosonVoucher.sol
```diff
@@ -279,4 +496,143 @@ contract BosonVoucher is IBosonVoucher, BeaconClientBase, OwnableUpgradeable, ER
...
+     * - does not yet have an owner
+     * - in a reserved range
+     * - has been pre-minted
+     * - has not been already burned
+     *
+     * @param _tokenId - the token id to check
+     * @return committable - whether the token is committable
+     * @return offerId - the associated offer id if committable
+     */
+    function getPreMintStatus(uint256 _tokenId) internal view returns (bool committable, uint256 offerId) {
```
anajuliabit, line 567: this functions looks great, pretty elegant and optimized way

contracts/protocol/clients/voucher/BosonVoucher.sol
zajck: Well, this still is a mint, it's just not done in OZ way. ERC721 defines only what contract should emit during the mint, but it does not require how contract stores the data.

So from the outside it looks completely normal (it emits the Transfer event, and if you query `ownerOf` tokenId it doesn't revert, it's possible to transfer it etc.). But since we are not using `_mint`, we save gas, since we don't make any checks done by it, and we save gas since we don't update `_owners` variable.

So we save at this point, which enables more efficient minting. But since there is no free lunch, this cost are paid at another time - when preminted token is transferred for the first time, this transfer is costlier compared to ordinary transfer. First, lookup to see if it's commitable is costlier than just checking if it exist + `_owners` gets populated, and putting something into an empty slot is always costlier than just updating the value.

contracts/protocol/clients/voucher/BosonVoucher.sol
zajck: FIxed (I used a slightly different wording)

contracts/protocol/clients/voucher/BosonVoucher.sol
zajck: Fixed

contracts/protocol/clients/voucher/BosonVoucher.sol
zajck: EVM storage does not carry any information about the type of data, stored in a certain slot. So, solidity compiler always relies on actual types defined in the code (for which we can say is a level above EVM).
Normally compiler assings each variable you define a certain slot and they don't overlap.

However, with assembly you can take an arbitrary storage slot and assign it any type you want. For example even if you define some variable of type address and store something there, you can always define another variable of another type (let's say uint) and point it to the same slot and whenever that is used in the code, compiler will treat is an uint.

Now in this exact case, it's probably the easiest to think about it this way:
- `_owners` and `_balances` are "true" variables. By "true" I mean variable defined on contract level, and their slots get assigned to them by compiler (and that's something we cannot manipulate)
- `ERC721UpgradeableStorage` is a virtual variable. It does not exist on the contract level and it's generated on the fly by calling `getERC721UpgradeableStorage`. So whenever that function is called we get only the storage pointer that points directly to the slot defined in it (which coincides with the `_owners` storage slot).

So why did I defined new struct and not simply do sth like
```
    function getOwnersStorage() internal pure returns (mapping(uint256 => address) storage ps) {
        assembly {
            ps.slot := 0x0000000000000000000000000000000000000000000000000000000000000099
        }
    }
```
and something similar for `_balances`. Well, I could do that, but I the time when I did it, I needed access to both `_owners` and `_balances` in the same function. So introducing a new struct `ERC721UpgradeableStorage` allowed me to do it in one call instead of two. This is possible since struct behave like contract level storage, i.e. their fields are packed in the slots in the same way as they would be if defined on contract level.
For example if struc has two fields, and let's say struct location is X, then first field is in the slot X and second in X+1

contracts/protocol/clients/voucher/BosonVoucher.sol
anajuliabit: Thank you for the explanation @zajck

contracts/protocol/clients/voucher/BosonVoucher.sol
anajuliabit: I was worried about overwriting old owners with the new data storage on ERC721UpgradeableStorage as is the same slot but that makes sense to me now, thanks.

anajuliabit (changes requested):
Could you please add the following tests:
- issueVoucher should revert if exchange id falls within a pre-minted offer's range
- Code coverage reported this branch as well but I'm not sure if we have a way to cover this
Screen Shot 2022-12-01 at 12 11 09
maybe calling ownerOf with a token id that doesn't exist and has no reserved range

cliffhall (commented):
LGTM, but it won't let me approve, since I'm a contributor on this PR. We may need to get @mischat or @levalleux-ludo to review and approve. Alternatively, when it's ready, I *think* I can force it through from the UI.

zajck (commented):
I added missing tests.

anajuliabit (approved):
LGTM

zajck:
Replaced by #483

PR #474, opened 2022-11-23
Script to upgrade client

zajck:
This PR introduces new script that:
- deploy new client (BosonVoucher)
- set new implementation address on beacon
- update addressess file

Script is tailored for the current situation where we have only one client.
If we add another client, this script will have to be updated (along with existing `deploy-protocol-clients`, `deploy-protocol-client-impls`, `deploy-protocol-client-proxies` and `deploy-protocol-client-beacons` which also assume that only one client exist.


docs/tasks.md
```diff
@@ -70,6 +70,23 @@ For upgrade to succeed you need an account with UPGRADER role. Refer to "Manage
 - **Ethereum Mainnet**. This upgrades the existing diamond on Ethereum Mainnet.  
 ```npm run upgrade-facets:ethereum:mainnet```
 
+### Upgrade facet
+Upgrade existing clients (currently only BosonVoucher). Script deploys new implementation and updates address on beacon.  
+We provide different npm scripts for different use cases. A script for Hardhat network does not exist. Since contracts are discarded after the deployment, they cannot be upgraded.  
+For upgrade to succeed you need an account with UPGRADER role. Refer to "Manage Roles" to see how to grant it.
```
anajuliabit, line 76:
```suggestion
For upgrade to succeed you need an account with UPGRADER role. Refer to [Manage roles](#manage-roles) to see how to grant it.
```

scripts/upgrade-clients.js
```diff
@@ -0,0 +1,101 @@
...
+ * - Admin must have UPGRADER role. Use `manage-roles.js` to grant it.
+ *
+ * Currently script upgrades the only existing client - BosonVoucher.
+ * If new clients are introduced, this script should be modified to get the list of clients to upgrade from the config.
+ */
+async function main(env) {
+  // Bail now if hardhat network, unless the upgrade is tested
+  if (network === "hardhat" && env !== "upgrade-test") process.exit();
+
+  const chainId = (await ethers.provider.getNetwork()).chainId;
```
anajuliabit, line 25: not a big deal but its easer to read

```suggestion
  const { chainId } = await ethers.provider.getNetwork();
```

scripts/upgrade-clients.js
```diff
@@ -0,0 +1,101 @@
...
+  const contractsFile = readContracts(chainId, network, env);
```
anajuliabit, line 26:
```suggestion
  const { contracts } = readContracts(chainId, network, env);
```

anajuliabit (changes requested):

docs/tasks.md
zajck: Done.

scripts/upgrade-clients.js
zajck: Done.

scripts/upgrade-clients.js
zajck: Done.

anajuliabit (approved):
LGTM

scripts/upgrade-clients.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  // Deploy Protocol Client implementation contracts
+  console.log(`\n📋 Deploying new logic contract`);
+  const [bosonVoucherImplementation] = await deployProtocolClientImpls(maxPriorityFeePerGas);
+
+  // Update implementation address on beacon contract
+  console.log(`\n📋 Updating implementation address on beacon`);
+  const beacon = await ethers.getContractAt("BosonClientBeacon", beaconAddress);
+  await beacon.setImplementation(bosonVoucherImplementation.address, await getFees(maxPriorityFeePerGas));
```
hswopeams, line 80: This line fails when I run on localhost with an error saying the caller doesn't have the correct role. 

`  const adminSigner = await ethers.getSigner(adminAddress);`

should be called first, and then `beacon` should connect to `adminSigner` when calling `setImplementation`

scripts/upgrade-clients.js
```diff
@@ -0,0 +1,100 @@
...
+
+  // Remove old entry from contracts
+  contracts = contracts.filter((i) => i.name !== "BosonVoucher Logic");
+  deploymentComplete("BosonVoucher Logic", bosonVoucherImplementation.address, [], "", contracts);
+
+  const contractsPath = await writeContracts(contracts, env);
```
hswopeams, line 86: I was just thinking that we have no way to version the client(s) separately from the protocol. What if we had a version where only the client changed and not the protocol? It's difficult because we only have on addresses file and one package.json file, which both contain the protocol version. I don't think that can be solved in this PR, but it's something to think about.

docs/tasks.md
```diff
@@ -70,6 +70,23 @@ For upgrade to succeed you need an account with UPGRADER role. Refer to "Manage
 - **Ethereum Mainnet**. This upgrades the existing diamond on Ethereum Mainnet.  
 ```npm run upgrade-facets:ethereum:mainnet```
 
+### Upgrade facet
```
hswopeams, line 73:
```suggestion
### Upgrade clients
```

hswopeams (changes requested):

docs/tasks.md
zajck: Fixed

scripts/upgrade-clients.js
zajck: Fixed.

scripts/upgrade-clients.js
zajck: Yes, probably we should keep separate versioning for clients.
Ideally these versions should also tell if client is compatible with protocol or not (for example when we release shell nft "client v2.2." will be compatible with all protocol versions, while "client v.2.1" won't be fully compatible with protocol v2.2

hswopeams (approved):

anajuliabit (approved):

PR #477, opened 2022-11-29
New voucher name and symbol

zajck:
Change voucher name in symbol to
- name: Boson Voucher (rNFT)
- symbol: BOSON_VOUCHER_RNFT


anajuliabit (changes requested):
Could you please also add (rNFT) to [docs/architeture.md#Voucher NFT](https://github.com/bosonprotocol/boson-protocol-contracts/blob/main/docs/architecture.md?plain=1#L48)

cliffhall:
I thought we were going to limit the term rNFT's incursion into the codebase by only changing the token name and symbol. This additional change of references in code and Ana's request to change the architecture.md go beyond that. This is the slippery slope. You have to draw a line somewhere and the further in you go with it, the harder it is to draw a clean line.

anajuliabit:
@cliffhall the changes that @zajck made on the docs are on places that we had called 'Voucher NFT' and 'rNFT better suits the current version of the protocol IMO

anajuliabit (dismissed):

zajck:
> I thought we were going to limit the term rNFT's incursion into the codebase by only changing the token name and symbol. This additional change of references in code and Ana's request to change the architecture.md go beyond that. This is the slippery slope. You have to draw a line somewhere and the further in you go with it, the harder it is to draw a clean line.

That;s what I did at first. Then after @anajuliabit's comment I made also some changes to code, but only where "NFT" already existed.

But now, giving it another though, I think that we should completely remove "NFT" or "rNFT" from the code and also from `docs/architecture.md`. My rationale:
- "Voucher" and "rNFT" are interchangeable, so it's enough to use only one of them (in our case voucher)
- Wherever "NFT" was still in the comments, it was more of a legacy thing. It may even lead to a confusion if voucher and NFT are two things.

We could say that "Voucher" represents the implementation. And "rNFT" is only a name that we use for our use case. If someone else deploys complete protocol, they are completely free to not use "rNFT". That's whay I would remove it everywhere.

anajuliabit:
@zajck You have a point. I agree!

anajuliabit (changes requested):
Please also remove from 
https://github.com/bosonprotocol/boson-protocol-contracts/blob/main/README.md?plain=1#L14
and
https://github.com/bosonprotocol/boson-protocol-contracts/blob/9d0f85fcb1fd3a84fe0419ab3d93617543012133/contracts/protocol/facets/SellerHandlerFacet.sol#L256

anajuliabit (approved):
LGTM

cliffhall (approved):
LGTM! 👍

PR #488, opened 2022-12-13
build boson-kleros integration poc

0xartem:
This is a draft for integration between kleros and boson protocol. The code is for information purposes and should NOT be merged directly to the main.

The following miro board contains some diagrams to inform the integration exploration.

https://miro.com/app/board/uXjVPet6BvQ=/

### DIFF
--- a/contracts/domain/BosonTypes.sol
+++ b/contracts/domain/BosonTypes.sol
@@ -193,6 +193,9 @@ contract BosonTypes {
     struct Dispute {
         uint256 exchangeId;
         uint256 buyerPercent;
+        uint256 sellerPercent;
+        bool buyerPercentSet;
+        bool sellerPercentSet;
         DisputeState state;
     }
 
--- a/contracts/interfaces/handlers/IBosonDisputeHandler.sol
+++ b/contracts/interfaces/handlers/IBosonDisputeHandler.sol
@@ -115,18 +115,9 @@ interface IBosonDisputeHandler is IBosonDisputeEvents, IBosonFundsLibEvents {
      * - Dispute was escalated and escalation period has elapsed
      *
      * @param _exchangeId  - the id of the associated exchange
-     * @param _buyerPercent - percentage of the pot that goes to the buyer
-     * @param _sigR - r part of the signer's signature
-     * @param _sigS - s part of the signer's signature
-     * @param _sigV - v part of the signer's signature
+     * @param _percent - percentage of the pot that goes to the buyer
      */
-    function resolveDispute(
-        uint256 _exchangeId,
-        uint256 _buyerPercent,
-        bytes32 _sigR,
-        bytes32 _sigS,
-        uint8 _sigV
-    ) external;
+    function resolveDispute(uint256 _exchangeId, uint256 _percent) external;
 
     /**
      * @notice Puts the dispute into the Escalated state.
@@ -211,14 +202,12 @@ interface IBosonDisputeHandler is IBosonDisputeEvents, IBosonFundsLibEvents {
      * @return dispute - the dispute details. See {BosonTypes.Dispute}
      * @return disputeDates - the dispute dates details {BosonTypes.DisputeDates}
      */
-    function getDispute(uint256 _exchangeId)
+    function getDispute(
+        uint256 _exchangeId
+    )
         external
         view
-        returns (
-            bool exists,
-            BosonTypes.Dispute memory dispute,
-            BosonTypes.DisputeDates memory disputeDates
-        );
+        returns (bool exists, BosonTypes.Dispute memory dispute, BosonTypes.DisputeDates memory disputeDates);
 
     /**
      * @notice Gets the state of a given dispute.
--- a/contracts/kleros/BosonKlerosConnector.sol
+++ b/contracts/kleros/BosonKlerosConnector.sol
@@ -1,0 +1,62 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8;
+// THIS CONTRACT IS IMPLEMENTED FOR INOFRMATION PURPOSES AND SHOULD NOT BE USED IN PRODUCTION
+
+import { IArbitrator } from "./IArbitrator.sol";
+import { IArbitrable } from "./IArbitrable.sol";
+import { IMetaEvidence } from "./IMetaEvidence.sol";
+import { IEscalatable } from "../interfaces/escalation/IEscalatable.sol";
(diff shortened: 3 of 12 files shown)


cliffhall:
> The following miro board contains some diagrams to inform the integration exploration.
> 
> https://miro.com/app/board/uXjVPet6BvQ=/

Since the Miro board is not public (I don't have access to it even with my Redeemeum account) would you mind posting a picture of it here and also summarize the approach taken by these changes? It would be nice to have a big picture understanding of what's going on before reviewing the PR.

0xartem:
> 

@cliffhall I am working on the BPIP. I will add the pictures there and a link to this PR.

contracts/protocol/facets/DisputeHandlerFacet.sol
```diff
@@ -226,20 +231,11 @@ contract DisputeHandlerFacet is DisputeBase, IBosonDisputeHandler {
...
+     * @param _percent - percentage of the pot that goes to the buyer or seller
      */
-    function resolveDispute(
-        uint256 _exchangeId,
-        uint256 _buyerPercent,
-        bytes32 _sigR,
-        bytes32 _sigS,
-        uint8 _sigV
-    ) external override disputesNotPaused nonReentrant {
+    function resolveDispute(uint256 _exchangeId, uint256 _percent) external override disputesNotPaused nonReentrant {
```
cliffhall, line 236: As mentioned in the [BPIP discussion](https://github.com/bosonprotocol/BPIPs/pull/1#issuecomment-1366135303), changing this method:

1) represents a breaking change that would require a major version rev to 3.0.0
2) discards the signature, which is required to ensure that the two parties are in accord.

It could be added as an alternative method with a name like `resolveDisputeWithProposal` or something, but it's not clear why this is even part of this proposed implementation. The BPIP is about using Kleros to handle escalations. This comes prior to escalation and is not a prerequisite for such.

contracts/domain/BosonTypes.sol
```diff
@@ -193,6 +193,9 @@ contract BosonTypes {
     struct Dispute {
         uint256 exchangeId;
         uint256 buyerPercent;
+        uint256 sellerPercent;
```
cliffhall, line 196: Changing the `Dispute` struct is dangerous. There are implications for existing storage.

1) The new fields are not at the end of the struct, and so previously stored structs will be misinterpreted and/or clobbered when written to.
2) Even if they were at the end of the struct, it isn't always safe to add members to structs, depending upon how those structs are stored, i.e., in arrays vs mappings.

For these reasons, we would more likely want to associate these new fields with the dispute via a mapping and not change the struct and all the client code (including the graph) that relies on its current shape.

contracts/interfaces/handlers/IBosonDisputeHandler.sol
```diff
@@ -115,18 +115,9 @@ interface IBosonDisputeHandler is IBosonDisputeEvents, IBosonFundsLibEvents {
...
+     * @param _percent - percentage of the pot that goes to the buyer
      */
-    function resolveDispute(
-        uint256 _exchangeId,
-        uint256 _buyerPercent,
-        bytes32 _sigR,
-        bytes32 _sigS,
-        uint8 _sigV
-    ) external;
+    function resolveDispute(uint256 _exchangeId, uint256 _percent) external;
```
cliffhall, line 120: See objections to this below on the `DisputeHandlerFacet`.

contracts/kleros/BosonKlerosConnector.sol
```diff
@@ -0,0 +1,62 @@
...
+pragma solidity ^0.8;
+// THIS CONTRACT IS IMPLEMENTED FOR INOFRMATION PURPOSES AND SHOULD NOT BE USED IN PRODUCTION
+
+import { IArbitrator } from "./IArbitrator.sol";
+import { IArbitrable } from "./IArbitrable.sol";
+import { IMetaEvidence } from "./IMetaEvidence.sol";
+import { IEscalatable } from "../interfaces/escalation/IEscalatable.sol";
+import { IEscalationResolver } from "../interfaces/escalation/IEscalationResolver.sol";
+
+contract BosonKlerosConnector is IEscalatable, IArbitrable, IMetaEvidence {
```
cliffhall, line 11: This should be organized under `protocol/clients/kleros/BosonKlerosConnector`. It also probably needs to be upgradeable.

contracts/kleros/BosonKlerosConnector.sol
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * @dev Give a ruling for a dispute. Must be called by the arbitrator.
+     * The purpose of this function is to ensure that the address calling it has the right to rule on the contract.
+     * @param _disputeID ID of the dispute in the Arbitrator contract.
+     * @param _ruling Ruling given by the arbitrator. Note that 0 is reserved for "Not able/wanting to make a decision".
+     */
+    function rule(uint256 _disputeID, uint256 _ruling) external {
+        BosonCase memory bosonCase = klerosDisputeCases[_disputeID];
+        if (bosonCase.exchangeId == 0) revert InvalidExchangeError();
```
cliffhall, line 49: Is the emission of an Error required by Kleros? Otherwise this would need to be implemented as a revert reason. We discussed implementing formal Error types but we were too far along the development track to do that before deploying V2, so we're kind of stuck with the style. 

We definitely shouldn't mix and match unless there's an absolute requirement.

That's said, it's in a client, which is a boundary contract that could _conceivably_ be given a pass if there's no other option.

cliffhall (commented):
Some comments about the implementation. Since it's not meant to be merged, I'm not requesting changes.

contracts/protocol/facets/DisputeHandlerFacet.sol
0xartem: Thanks for the comment @cliffhall. I am going to update the BPIP so we don't introduce the breaking change.

A new method is needed because Kleros requires two options to choose from. It means that escalation only makes sense when a buyer and a seller have proposals that don't match, like 50% and 70%.

A state of those proposals must be stored somewhere before the buyer escalates it to a DR.

contracts/domain/BosonTypes.sol
0xartem: Noted, will change it in the new PR.

contracts/kleros/BosonKlerosConnector.sol
0xartem: There is no such requirement from Kleros. It can be modified to match Boson coding style.

PR #491, opened 2022-12-16
Add pre-commit hook

anajuliabit:
Add pre-commit using husky lib to run `tidy:scripts`, `tidy:contracts`, and `natspec-interface-id:fix`.

TODO:
- [x] Only run `tidy:scripts` when changing javascript files
- [x] Only run `tidy:contracts` when changing contracts


package.json
```diff
@@ -66,7 +66,8 @@
     "manage-roles:test": "npx hardhat manage-roles --network test --env test >> logs/test.manage.roles.txt",
     "natspec-interface-id": "npx hardhat compile && node scripts/util/natspec-interfaceid.js",
     "natspec-interface-id:fix": "npx hardhat compile && node scripts/util/natspec-interfaceid.js --fix",
-    "estimate-limits": "npx hardhat clean && npx hardhat compile && npx hardhat estimate-limits"
+    "estimate-limits": "npx hardhat clean && npx hardhat compile && npx hardhat estimate-limits",
+    "prepare": "husky install"
```
cliffhall, line 70: I realize the husky docs say make a one-one time script called `prepare` to get going.

However, in the context of our repo "prepare" doesn't give much context. And if you don't know what husky is for, most likely you will not run this. Because there is no documentation explaining what it is for and when / why you should run it.

Since it is to be run once at developer setup time and it needs documentation, I suggest removing the script from this file entirely and instead adding a section called "**Prepare pre-commit hooks**" in [local-development.md](https://github.com/bosonprotocol/boson-protocol-contracts/blob/main/docs/local-development.md) between the [Install required Node modules](https://github.com/bosonprotocol/boson-protocol-contracts/blob/main/docs/local-development.md#install-required-node-modules) and [Configure Environment](https://github.com/bosonprotocol/boson-protocol-contracts/blob/main/docs/local-development.md#configure-environment) sections. It should just explain what pre-commit hooks are defined and that they are installed with `husky install`

cliffhall (changes requested):
Suggest we just use the developer setup doc instead of an npm script for installing the hook.

package.json
anajuliabit: @cliffhall Since NPM v4,`prepare` script is executed when you run npm install

> Runs BEFORE the package is packed, i.e. during npm publish and npm pack
> Runs on local npm install without any arguments
> Runs AFTER prepublish, but BEFORE prepublishOnly

Check https://docs.npmjs.com/cli/v9/using-npm/scripts

cliffhall: Right. We don't need it to run this every time we run npm install though. Only the first time.

anajuliabit: changed and documentation added

docs/local-development.md
```diff
@@ -29,6 +29,18 @@ cd boson-protocol-contracts
...
+### Prepare git pre-hooks
+Pre-hooks are scripts that runs automatically before each [git hook](https://git-scm.com/docs/githooks) execution. 
+We use the pre-commit hook to ensure that all new code follows the style guidelines defined in [eslint](../.eslintrc.json) and [prettier](../.prettierrc) configuration files before it is committed to the repository.
+The pre-commit hook is also used to verify and fix [natspec interfaces ids](./tasks.md#verify-natspec-interface-ids)
+
+The [Husky](https://typicode.github.io/husky) library is used to manage pre-hooks scripts.
+To make husky work, you must run:
+
+```
+husky install
```
zajck, line 41: That didn't work for me, I had to run it with

```npx husky install```

anajuliabit: fixed

zajck (approved):

cliffhall (approved):
LGTM! 👍

PR #497, opened 2022-12-22
Fix detect contract change

zajck:
Script to detect which contracts were changed, didn't work properly if some contract was removed.
This PR fixes it.


scripts/util/detect-changed-contracts.js
```diff
@@ -100,7 +106,9 @@ async function detectChangedContract(referenceCommit, targetCommit) {
   console.log(`Total changed contracts: ${changedContracts.length}`);
 
   // Checkout back to original branch
+  shell.exec(`rm -rf contracts/*`);
```
cliffhall, line 109: Doesn't need the glob, but if leaving in, should add into line 53 which does the same without the glob.

cliffhall (changes requested):

scripts/util/detect-changed-contracts.js
zajck: Fixed.

cliffhall (approved):
LGTM! 👍

PR #499, opened 2023-01-04
feat: return deployed contract addresses

albertfolch-redeemeum:
This branch has been created from the tag v2.2.0-rc.1-test and you can see the diff in https://github.com/bosonprotocol/boson-protocol-contracts/compare/v2.2.0-rc.1-test...set-nth-auth-contracts-addresses?expand=1

These changes are needed so that on the core-components side we can be sure that the authTokenContract addresses are set properly. As we use your deploy functions to deploy the contracts locally, we could end up with different contract addresses (which is the current issue this PR helps to fix).


mischat:
nice to see you here @albertfolch-redeemeum :)

scripts/deploy-suite.js
```diff
@@ -158,7 +158,7 @@ async function main(env, facetConfig) {
   // LENS
   transactionResponse = await bosonConfigHandler.setAuthTokenContract(
     AuthTokenType.Lens,
-    authTokenContracts.lensAddress,
+    process.env.LENS_ADDRESS || authTokenContracts.lensAddress,
```
zajck, line 161: Can you please make this change in function `getAuthTokenContracts()` (at the top of the file) instead of here?
And the same goes for `process.env.ENS_ADDRESS`.

zajck (changes requested):
Beside other requests, please also add `LENS_ADDRESS` and `ENS_ADDRESS` to `.env.example`

scripts/deploy-suite.js
albertfolch-redeemeum: changed, thanks!

zajck (approved):

levalleux-ludo (approved):

PR #515, opened 2023-01-05
[PIF-02M] Unprotected Facet w/ Delegate Call

### LINKED ISSUE #507: [PIF-02M] Unprotected Facet w/ Delegate Call
## PIF-02M: Unprotected Facet w/ Delegate Call

| [Language Specific](/appendix/finding-types#language-specific) |  | ProtocolInitializationFacet.sol:L54 |

### Description:

The `ProtocolInitializationFacet` can perform `delegatecall` instructions and represents a logic implementation to which calls should be relayed to via the Diamond implementation. As the base implementation deployed for the Diamond to point to is unprotected, a malicious party can cause it to self-destruct rendering the `getVersion` function un-invoke-able.

### Impact:

By self-destructing the version initializer, it would become impossible to query the current version via `getVersion` which future modules of Boson could rely on (such as automated version increment systems). Additionally, the module would be unusable in future version upgrades of the system requiring a re-deployment of the initializer.

### Example:

```solidity
/**
 * @notice Initializes the protocol after the deployment.
 * This function is callable only once for each version
 *
 * @param _version - version of the protocol
 * @param _addresses - array of facet addresses to call initialize methods
 * @param _calldata -  array of facets initialize methods encoded as calldata
 *                    _calldata order must match _addresses order
 * @param _isUpgrade - flag to indicate whether this is first deployment or upgrade
 * @param _initializationData - data for initialization of the protocol, using this facet (only if _isUpgrade == true)
 * @param _interfacesToRemove - array of interfaces to remove from the diamond
 * @param _interfacesToAdd - array of interfaces to add to the diamond
 */
function initialize(
    bytes32 _version,
    address[] calldata _addresses,
    bytes[] calldata _calldata,
    bool _isUpgrade,
    bytes calldata _initializationData,
    bytes4[] calldata _interfacesToRemove,
    bytes4[] calldata _interfacesToAdd
) external onlyUninitializedVersion(_version) {
    require(_version != bytes32(0), VERSION_MUST_BE_SET);
    require(_addresses.length == _calldata.length, ADDRESSES_AND_CALLDATA_LENGTH_MISMATCH);

    // Delegate call to initialize methods of facets declared in _addresses
    for (uint256 i = 0; i < _addresses.length; i++) {
        (bool success, bytes memory error) = _addresses[i].delegatecall(_calldata[i]);

        // Handle result
        if (!success) {
            if (error.length > 0) {
                // bubble up the error
                revert(string(error));
            } else {
                // Reverts with default message
                revert(PROTOCOL_INITIALIZATION_FAILED);
            }
        }
    }

    ProtocolLib.ProtocolStatus storage status = protocolStatus();
    if (_isUpgrade) {
        if (_version == bytes32("2.2.0")) {
            initV2_2_0(_initializationData);
        }
    }

    removeInterfaces(_interfacesToRemove);
    addInterfaces(_interfacesToAdd);

    status.version = _version;
    emit ProtocolInitialized(string(abi.encodePacked(_version)));
}
```

### Recommendation:

We advise a `constructor` to be introduced that sets an `immutable` variable with the value of `address(this)`. As a next step, we advise a `require` check to be introduced to `initialize` that ensures only proxies invoke it by validating that `address(this) != immVar` where `immVar` represents the immutable variable we declared and assigned to in the previous step. This protection measure will prevent the `initialize` function from being invoked arbitrarily on the base implementation and thus prevent it from ever being self-destruct.

zajck:
Fixes #507

### DIFF
--- a/contracts/protocol/facets/ProtocolInitializationHandlerFacet.sol
+++ b/contracts/protocol/facets/ProtocolInitializationHandlerFacet.sol
@@ -14,6 +14,8 @@ import { DiamondLib } from "../../diamond/DiamondLib.sol";
  *
  */
 contract ProtocolInitializationHandlerFacet is IBosonProtocolInitializationHandler, ProtocolBase {
+    address private thisAddress; // used to prevent invocation of initialize directly on deployed contract. Variable is not used by the protocol.
+
     /**
      * @notice Modifier to protect initializer function from being invoked twice for a given version.
      */
@@ -22,6 +24,15 @@ contract ProtocolInitializationHandlerFacet is IBosonProtocolInitializationHandl
         require(!ps.initializedVersions[_version], ALREADY_INITIALIZED);
         ps.initializedVersions[_version] = true;
         _;
+    }
+
+    /**
+     * @notice Constructor
+     *
+     * @dev This constructor is used to prevent invocation of initialize directly on deployed contract.
+     */
+    constructor() {
+        thisAddress = address(this);
     }
 
     /**
@@ -46,6 +57,7 @@ contract ProtocolInitializationHandlerFacet is IBosonProtocolInitializationHandl
         bytes4[] calldata _interfacesToRemove,
         bytes4[] calldata _interfacesToAdd
     ) external onlyUninitializedVersion(_version) {
+        require(address(this) != thisAddress, DIRECT_INITIALIZATION_NOT_ALLOWED);
         require(_version != bytes32(0), VERSION_MUST_BE_SET);
         require(_addresses.length == _calldata.length, ADDRESSES_AND_CALLDATA_LENGTH_MISMATCH);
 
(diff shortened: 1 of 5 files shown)


contracts/protocol/facets/ProtocolInitializationHandlerFacet.sol
```diff
@@ -24,6 +26,15 @@ contract ProtocolInitializationHandlerFacet is IBosonProtocolInitializationHandl
...
+     *
+     * @dev This constructor is used to prevent invocation of initialize directly on deployed contract.
+     */
+    constructor() {
+        thisAddress = address(this);
+    }
+
     /**
      * @notice Initializes the protocol after the deployment.
      * This function is callable only once for each version
```
mischat, line 40: I have noticed that in some other places we have put the conditions in the comments, this is an external method, worth doing it here ?

mischat (approved):
I liked the test you wrote there to show the selfdestruct in action ...

contracts/protocol/facets/ProtocolInitializationHandlerFacet.sol
zajck: I added them.

Maybe initialize are slightly different than other external methods, since they are not part of the interface, and they don't go into documentation. But it make sense to put the conditions here for other devs.

anajuliabit (approved):
LGTM

PR #516, opened 2023-01-05
fix: getInterfaceNames so that it works in core-components

albertfolch-redeemeum:
as this repo is added in core-components as a submodule in another subfolder called _protocol-contracts_, our deploy script fails. This should fix it

### DIFF
--- a/scripts/config/supported-interfaces.js
+++ b/scripts/config/supported-interfaces.js
@@ -69,7 +69,10 @@ async function getInterfaceNames() {
     const [source, name] = contractName.split(":");
 
     // If starts with prefix and is not in skip list, return name
-    return /.*contracts\/interfaces\/(.*)/.test(source) && !skip.some((s) => new RegExp(`.*contracts/interfaces/${s}`).test(source)) ? name : [];
+    return /.*contracts\/interfaces\/(.*)/.test(source) &&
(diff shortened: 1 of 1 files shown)


zajck (commented):
I'm wondering - can this be done more flexible?

What I mean is, if somebody just clones contract repo, `prefix` sort of makes sense to them, while  `prefixProtocolContracts` makes sense to them only if they are aware of core-component repo.
Or if we look from another perspective - if someone else starts using protocol contracts as submodule in their repo, they would need to come here and add their prefix. 

Maybe there is a more general solution than adding multiple prefixes? :thinking:

levalleux-ludo:
> I'm wondering - can this be done more flexible?
> 
> What I mean is, if somebody just clones contract repo, `prefix` sort of makes sense to them, while `prefixProtocolContracts` makes sense to them only if they are aware of core-component repo. Or if we look from another perspective - if someone else starts using protocol contracts as submodule in their repo, they would need to come here and add their prefix.
> 
> Maybe there is a more general solution than adding multiple prefixes? 🤔

Agree with Klemen

levalleux-ludo (changes requested):
see comments from Klemen and myself

albertfolch-redeemeum:
changed @levalleux-ludo @zajck

zajck (approved):
LGTM :+1:

levalleux-ludo (approved):

PR #523, opened 2023-01-06
[EHF-01M] Potential Fatal Failure of Voucher Redemption

### LINKED ISSUE #522: [EHF-01M] Potential Fatal Failure of Voucher Redemption
## EHF-01M: Potential Fatal Failure of Voucher Redemption

| [Logical Fault](/appendix/finding-types#logical-fault) |  | ExchangeHandlerFacet.sol:L448, [L471](https://github.com/bosonprotocol/boson-protocol-contracts/blob/38c7ba384e55486aee4ea76bb80b4f29d1a86175/contracts/protocol/facets/ExchangeHandlerFacet.sol#L471), [L810](https://github.com/bosonprotocol/boson-protocol-contracts/blob/38c7ba384e55486aee4ea76bb80b4f29d1a86175/contracts/protocol/facets/ExchangeHandlerFacet.sol#L810) |

### Description:

The `redeemVoucher` function will invoke the `transferTwins` function which in turn will cause a dispute to be automatically raised if the sender is a contract via `raiseDisputeInternal`. Due to a revamp of the function in `DisputeBase`, it now will fail if the disputes region is paused (`disputesNotPaused` modifier). As a result, voucher redemption by smart contracts that could fail will be impossible to execute in case the caller is a contract.

### Impact:

Due to the new changes introduced in the dispute raising mechanism, a smart contract which was previously coded with an assumption that its voucher redemption call will succeed when the dispute region is paused will now fail. Such an unhandled failure can lead to unredeemable vouchers as well as general misbehaviour at the integration level of the caller. It is important that upgrades to the protocol do not break expected behaviour of externally-integrating modules to ensure that integrations of Boson Protocol can properly equip their contracts without needing built-in upgradeable capabilities.

### Example:

```solidity
if (transferFailed) {
    // Raise a dispute if caller is a contract
    if (sender.isContract()) {
        raiseDisputeInternal(_exchange, _voucher, seller.id);
    } else {
        // Revoke voucher if caller is an EOA
        revokeVoucherInternal(_exchange);
        // N.B.: If voucher was revoked because transfer twin failed, then voucher was already burned
        shouldBurnVoucher = false;
    }
}
```

### Recommendation:

We advise the error to be handled gracefully via a `try-catch` structure that will cause the voucher revocation mechanism to kick in if a dispute cannot be raised due to the said region being paused.

zajck:
Fixes #522

### DIFF
--- a/contracts/protocol/bases/DisputeBase.sol
+++ b/contracts/protocol/bases/DisputeBase.sol
@@ -17,7 +17,6 @@ contract DisputeBase is ProtocolBase, IBosonDisputeEvents {
      * @notice Raises a dispute
      *
      * Reverts if:
-     * - The disputes region of protocol is paused
      * - Caller does not hold a voucher for the given exchange id
      * - Exchange does not exist
      * - Dispute period has elapsed already
@@ -30,7 +29,7 @@ contract DisputeBase is ProtocolBase, IBosonDisputeEvents {
         Exchange storage _exchange,
         Voucher storage _voucher,
         uint256 _sellerId
-    ) internal disputesNotPaused {
+    ) internal {
         // Fetch offer durations
         OfferDurations storage offerDurations = fetchOfferDurations(_exchange.offerId);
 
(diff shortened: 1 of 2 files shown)


anajuliabit (approved):
this implementation is different from their suggestion, it would be nice to make it explicitly in the issue

zajck:
> this implementation is different from their suggestion, it would be nice to make it explicitly in the issue

@anajuliabit I added as comment in referenced issue.

mischat (approved):
👍

PR #526, opened 2023-01-10
fix: INVALID_RANGE_LENGTH error text

albertfolch-redeemeum:


zajck (changes requested):
Good catch.

Please also change it in `scripts/config/revert-reasons.js`, otherwise unit tests will fail.

zajck (approved):

anajuliabit (approved):

PR #527, opened 2023-01-11
Create dependabot.yml

mischat:
I think this is the right config, please do suggest if we need something else @zajck or @anajuliabit


zajck (approved):
LGTM

anajuliabit (approved):
LGTM

PR #539, opened 2023-01-16
Remove files before the build

zajck:
Simple fix that prevent the failure of doc generation in https://github.com/bosonprotocol/docs.bosonprotocol.io
In some cases if there some old files remain the the doc folder, it can result in broken link. This PR solves that by adding a command that always cleans the folder with old docs before generating new.


anajuliabit (approved):

mischat (approved):
seems sensible

PR #541, opened 2023-01-18
Initial config to run slither analyzer on CI

anajuliabit:


slither.config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "filter_paths": "(contracts/mock/|contracts/ext_libs/|node_modules)"
```
zajck, line 2: I suggest we exclude detectors:
- `uninitialized-local` since we intentionally never initialize default values.
- `timestamp` since we want comparison agains external timestamp and miner's potentioal manipulation has low to no impact
- `assembly` since we use it as little as possible, but when we do, it's intentional (and it's not just some random copy paste). And everything that this detector does it just says "don't use assembly, it's bad", but does not point out actual problems
-

.github/workflows/ci.yaml
```diff
@@ -125,3 +125,39 @@ jobs:
...
+        uses: actions/checkout@v3.1.0
+      - name: Setup node
+        uses: actions/setup-node@v3.5.1
+        with:
+          node-version: 16.14.x
+          cache: 'npm'
+      - name: Install Dependencies
+        run: npm install
+      # - name: Compile Contracts
+        # run: npm run build
```
zajck, line 145: If this is not needed, just remove it completely.

zajck (changes requested):

.github/workflows/ci.yaml
anajuliabit: fixed

slither.config.json
anajuliabit: I’ve excluded the detectors you mentioned

zajck (dismissed):
LGTM.

mischat (commented):
This looks fine to me. Is the intention to ship this into v2.2 ? 

Last time we spoke about this we kept the lower version number due to it being what hardhat used and what omniscia used for their testing. 

Just to reiterate, i would personally always take the latest version of solidity...

anajuliabit:
@mischat hardhat now supports versions up to 0.8.17 https://hardhat.org/hardhat-runner/docs/reference/solidity-support
The latest is 0.8.17 and slither recommends not using the latest, but one below (0.8.16)

mischat:
@anajuliabit and @zajck given that omniscia used the 0.8.9. and the only "fix" here seems to be the "immutable" added in the snapshotgate, i think i would be happy for this to come in right after we release v2.2. 

Happy to discuss if need be, that way the auditors would be happy that we are releasing against the same version they tested the code against. I would be happy to not add this to the v2.2 release and then to add it in right after we release that ?

anajuliabit:
I'm happy about postponing the version upgrade. If @zajck agrees I can undo the version upgrade so we can merge this PR.

zajck (approved):

PR #552, opened 2023-01-25
chore: add forwarder address in deploy suite script

albertfolch-redeemeum:
in order to be able to specify a forwarder address after it has been deployed and to override the dummy values, I added this env variable


levalleux-ludo (approved):

zajck (changes requested):
Can you please add `FORWARDER_ADDRESS` to `.env.example` as well

zajck (approved):

levalleux-ludo (approved):

PR #554, opened 2023-01-26
Upgrade clients to v2.2.0-rc.2 on staging and test env

anajuliabit:
This PR adds logs of upgrade to tag `v2.2.0-rc.2` on staging and test environments.

This PR also fixes a bug on upgrade-clients script.


zajck (commented):
Given that only clients were updated, we cannot check in addresses file with new protocol version and old protocol contract addresses.

PR #557, opened 2023-02-02
Fix bug on tokenURI method on BosonVoucher

anajuliabit:
TokenURI was returning data only for offers with existing exchanges. 
Preminted vouchers have no exchange associated.

### DIFF
--- a/contracts/protocol/clients/voucher/BosonVoucher.sol
+++ b/contracts/protocol/clients/voucher/BosonVoucher.sol
@@ -499,16 +499,24 @@ contract BosonVoucherBase is IBosonVoucher, BeaconClientBase, OwnableUpgradeable
      * replaceable baseURI template, since the latter is not compatible
      * with IPFS hashes.
      *
-     * @param _exchangeId - id of the voucher's associated exchange
+     * @param _tokenId - id of the voucher's associated exchange or pre-minted
      * @return the uri for the associated offer's off-chain metadata (blank if not found)
      */
-    function tokenURI(uint256 _exchangeId)
+    function tokenURI(uint256 _tokenId)
         public
         view
         override(ERC721Upgradeable, IERC721MetadataUpgradeable)
         returns (string memory)
     {
-        (bool exists, Offer memory offer) = getBosonOfferByExchangeId(_exchangeId);
+        (bool exists, Offer memory offer) = getBosonOfferByExchangeId(_tokenId);
+
+        if (!exists) {
+            (bool committable, uint256 offerId) = getPreMintStatus(_tokenId);
+            if (committable) {
+                exists = true;
+                (offer, ) = getBosonOffer(offerId);
+            }
+        }
         return exists ? offer.metadataUri : "";
     }
 
--- a/contracts/protocol/clients/voucher/BosonVoucher.sol
+++ b/contracts/protocol/clients/voucher/BosonVoucher.sol
@@ -499,7 +499,7 @@ contract BosonVoucherBase is IBosonVoucher, BeaconClientBase, OwnableUpgradeable
      * replaceable baseURI template, since the latter is not compatible
      * with IPFS hashes.
      *
-     * @param _tokenId - id of the voucher's associated exchange or pre-minted
+     * @param _tokenId - id of the voucher's associated exchange or pre-minted token id
      * @return the uri for the associated offer's off-chain metadata (blank if not found)
      */
     function tokenURI(uint256 _tokenId)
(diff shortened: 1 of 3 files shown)


contracts/protocol/clients/voucher/BosonVoucher.sol
```diff
@@ -499,16 +499,24 @@ contract BosonVoucherBase is IBosonVoucher, BeaconClientBase, OwnableUpgradeable
      * replaceable baseURI template, since the latter is not compatible
      * with IPFS hashes.
      *
-     * @param _exchangeId - id of the voucher's associated exchange
+     * @param _tokenId - id of the voucher's associated exchange or pre-minted
```
zajck, line 502:
```suggestion
     * @param _tokenId - id of the voucher's associated exchange or pre-minted token id
```

test/protocol/clients/BosonVoucherTest.js
```diff
@@ -1861,6 +1865,19 @@ describe("IBosonVoucher", function () {
       const tokenURI = await bosonVoucher.tokenURI(1);
       expect(tokenURI).eq(metadataUri);
     });
+
+    it("should return the correct tokenURI when token is preminted", async function () {
```
zajck, line 1869: Can you please add another test, where token is preminted and then transferred. Just to make sure it then behaves as normal and get the tokenURI.

zajck (changes requested):
Just a note - I actually checked ERC721 standard again and it states that `tokenURI` "Throws if `_tokenId` is not a valid NFT". So we are not strictly compliant with the standard (this was the case already before this PR). 
One thing is that it returns empty string when it should revert, but I'm ok with that.
The other thing is that it if one supplies `_tokenId` of already redeemed voucher, it will return the metadata even though NFT does not exist anymore. This is also probably not problematic, since noone should rely on tokenURI to determine existance of the voucher.
I don't think we need to make any changes, but I'm just pointing this out, so if  anyone else think we should be fully compliant, it would be good to do it now.

test/protocol/clients/BosonVoucherTest.js
anajuliabit: added

contracts/protocol/clients/voucher/BosonVoucher.sol
anajuliabit: dode

zajck (approved):
LGTM

levalleux-ludo (approved):
LGTM

mischat:
> Just a note - I actually checked ERC721 standard again and it states that `tokenURI` "Throws if `_tokenId` is not a valid NFT". So we are not strictly compliant with the standard (this was the case already before this PR). One thing is that it returns empty string when it should revert, but I'm ok with that. The other thing is that it if one supplies `_tokenId` of already redeemed voucher, it will return the metadata even though NFT does not exist anymore. This is also probably not problematic, since noone should rely on tokenURI to determine existance of the voucher. I don't think we need to make any changes, but I'm just pointing this out, so if anyone else think we should be fully compliant, it would be good to do it now.

I guess, the thought that our preMinted voucher is not a valid NFT depends on the definition of a valid NFT :) I would agree that it is a valid NFT :) I think our pre-minted voucher seems like a "valid nft" to me, it is transferable right? Anyways, i guess i will look up the definition of a valid NFT....

I think your point re: post redemption, and the fact that the voucher contract still returns a tokenURI for a burned token, seems like something we might want to look at in the future ... Perhaps we should log an issue highlighting this. I also agree that people shouldn't be using the tokenURI getter to check the existence of an NFT, but they might end up doing that and we do break the spec here don't we ?

mischat:
> > Just a note - I actually checked ERC721 standard again and it states that `tokenURI` "Throws if `_tokenId` is not a valid NFT". So we are not strictly compliant with the standard (this was the case already before this PR). One thing is that it returns empty string when it should revert, but I'm ok with that. The other thing is that it if one supplies `_tokenId` of already redeemed voucher, it will return the metadata even though NFT does not exist anymore. This is also probably not problematic, since noone should rely on tokenURI to determine existance of the voucher. I don't think we need to make any changes, but I'm just pointing this out, so if anyone else think we should be fully compliant, it would be good to do it now.
> 
> I guess, the thought that our preMinted voucher is not a valid NFT depends on the definition of a valid NFT :) I would agree that it is a valid NFT :) I think our pre-minted voucher seems like a "valid nft" to me, it is transferable right? Anyways, i guess i will look up the definition of a valid NFT....
> 
> I think your point re: post redemption, and the fact that the voucher contract still returns a tokenURI for a burned token, seems like something we might want to look at in the future ... Perhaps we should log an issue highlighting this. I also agree that people shouldn't be using the tokenURI getter to check the existence of an NFT, but they might end up doing that and we do break the spec here don't we ?

Ah, so, looking at the code we don't throw, we return an empty string `""` is this what you meant by "we are not standard compliant" ? If so, this seems like an easy enough fix...

mischat (commented):
This LGTM too

PR #498, opened 2022-12-29
Add tests upgrade to v2.2.0 

anajuliabit:
This PR adds integration tests to insure protocol integrity on upgrade from 2.1.0 to 2.2.0

## Breaking changes
-   **PR**: Allows DR creation without the need for activation by protocol admin by @anajuliabit in https://github.com/bosonprotocol/boson-protocol-contracts/pull/466
      - [x] DR can be activated on creation.
- **PR**: Allowlist metatx methods by @zajck in https://github.com/bosonprotocol/boson-protocol-contracts/pull/461
   - [x] Meta transaction should work with allowlisted function
   - [x] Meta transaction should fail when function name is not allowlisted
   - [x] functionPointer positions should change
- **PR**: Raise and Escalate Dispute Orchestration by @cliffhall in https://github.com/bosonprotocol/boson-protocol-contracts/pull/489:
   - [x] Test if facets address pointed to selectors from OrchestrationHandlerFacet were replaced for facets OrchestrationHandlerFacet1 and OrchestrationHandlerFacet2
- **PR**: New voucher name and symbol by @zajck in https://github.com/bosonprotocol/boson-protocol-contracts/pull/477
  - [x] Test if new vouchers are created with new name and symbol
- **PR**: Rename operator to assistant #556
  - [x] Operator has been renamed to assistant
 
   
## New methods
- **PR**: Allowlist metatx methods by @zajck in https://github.com/bosonprotocol/boson-protocol-contracts/pull/461
   - [x] setAllowlistedFunctions
   - [x] isFunctionAllowlisted(bytes32)
   - [x] isFunctionAllowlisted(string)
   - [x] Meta transaction should fail when function name is not allowlisted
- **PR**: Add new facet ProtocolInitializationFacet by @anajuliabit in https://github.com/bosonprotocol/boson-protocol-contracts/pull/475
   - [x] getVersion
- **PR**: Init v2_2_0 by @zajck in https://github.com/bosonprotocol/boson-protocol-contracts/pull/492
   - [x] Should call initV2_2_0
- **PR**: Raise and Escalate Dispute Orchestration by @cliffhall in https://github.com/bosonprotocol/boson-protocol-contracts/pull/489
   - [x] raiseAndEscalateDispute
 - **PR**: Protocol support preminted voucher by @zajck in https://github.com/bosonprotocol/boson-protocol-contracts/pull/490
     - [x] setMaxPremintedVouchers
     - [x] getMaxPremintedVouchers
[...]

PR #563, opened 2023-02-09
Allow vouchers to be preMinted to voucher contract

anajuliabit:
- Vouchers can now be pre-minted for both the contract owner and contract address. 
- A new parameter `_to` has been added to `reserveRange` and must be either the contract owner or contract address. 
-  I have also fixed the pre-commit hook script in this PR.
 
## Notes
 I'll add the upgrade tests checking if OfferHandler has new changes in this [PR](https://github.com/bosonprotocol/boson-protocol-contracts/pull/498) after we merge this one to main, in order to avoid conflicts

### DIFF
--- a/contracts/interfaces/handlers/IBosonOrchestrationHandler.sol
+++ b/contracts/interfaces/handlers/IBosonOrchestrationHandler.sol
@@ -13,7 +13,7 @@ import { IBosonBundleEvents } from "../events/IBosonBundleEvents.sol";
  *
  * @notice Combines creation of multiple entities (accounts, offers, groups, twins, bundles) in a single transaction
  *
- * The ERC-165 identifier for this interface is: 0x6fa524ec
+ * The ERC-165 identifier for this interface is: 0xa38bc2e7
  */
 interface IBosonOrchestrationHandler is
     IBosonAccountEvents,
@@ -162,6 +162,7 @@ interface IBosonOrchestrationHandler is
      * - When agent id is non zero:
      *   - If Agent does not exist
      *   - If the sum of agent fee amount and protocol fee amount is greater than the offer fee limit
+     * - _to is not the BosonVoucher contract address or the BosonVoucher contract owner
      *
      * @dev No reentrancy guard here since already implemented by called functions. If added here, they would clash.
      *
@@ -171,6 +172,7 @@ interface IBosonOrchestrationHandler is
      * @param _offerDurations - the fully populated offer durations struct
      * @param _disputeResolverId - the id of chosen dispute resolver (can be 0)
      * @param _reservedRangeLength - the amount of tokens to be reserved for preminting
+     * @param _to - the address to send the pre-minted vouchers to (contract address or contract owner)
      * @param _authToken - optional AuthToken struct that specifies an AuthToken type and tokenId that the seller can use to do admin functions
      * @param _voucherInitValues - the fully populated BosonTypes.VoucherInitValues struct
      * @param _agentId - the id of agent
@@ -182,6 +184,7 @@ interface IBosonOrchestrationHandler is
         BosonTypes.OfferDurations calldata _offerDurations,
         uint256 _disputeResolverId,
         uint256 _reservedRangeLength,
+        address _to,
         BosonTypes.AuthToken calldata _authToken,
         BosonTypes.VoucherInitValues calldata _voucherInitValues,
         uint256 _agentId
@@ -268,6 +271,7 @@ interface IBosonOrchestrationHandler is
      * - When agent id is non zero:
      *   - If Agent does not exist
      *   - If the sum of agent fee amount and protocol fee amount is greater than the offer fee limit
+     * - _to is not the BosonVoucher contract address or the BosonVoucher contract owner
      *
      * @dev No reentrancy guard here since already implemented by called functions. If added here, they would clash.
      *
@@ -276,6 +280,7 @@ interface IBosonOrchestrationHandler is
      * @param _offerDurations - the fully populated offer durations struct
      * @param _disputeResolverId - the id of chosen dispute resolver (can be 0)
      * @param _reservedRangeLength - the amount of tokens to be reserved for preminting
+     * @param _to - the address to send the pre-minted vouchers to (contract address or contract owner)
      * @param _condition - the fully populated condition struct
      * @param _agentId - the id of agent
      */
@@ -285,6 +290,7 @@ interface IBosonOrchestrationHandler is
         BosonTypes.OfferDurations calldata _offerDurations,
         uint256 _disputeResolverId,
         uint256 _reservedRangeLength,
+        address _to,
         BosonTypes.Condition calldata _condition,
         uint256 _agentId
     ) external;
@@ -376,6 +382,7 @@ interface IBosonOrchestrationHandler is
      * - When agent id is non zero:
      *   - If Agent does not exist
      *   - If the sum of agent fee amount and protocol fee amount is greater than the offer fee limit
+     * - _to is not the BosonVoucher contract address or the BosonVoucher contract owner
      *
      * @dev No reentrancy guard here since already implemented by called functions. If added here, they would clash.
      *
@@ -384,6 +391,7 @@ interface IBosonOrchestrationHandler is
      * @param _offerDurations - the fully populated offer durations struct
      * @param _disputeResolverId - the id of chosen dispute resolver (can be 0)
      * @param _reservedRangeLength - the amount of tokens to be reserved for preminting
+     * @param _to - the address to send the pre-minted vouchers to (contract address or contract owner)
      * @param _groupId - id of the group, to which offer will be added
      * @param _agentId - the id of agent
      */
@@ -393,6 +401,7 @@ interface IBosonOrchestrationHandler is
         BosonTypes.OfferDurations calldata _offerDurations,
         uint256 _disputeResolverId,
         uint256 _reservedRangeLength,
+        address _to,
         uint256 _groupId,
         uint256 _agentId
     ) external;
@@ -494,6 +503,7 @@ interface IBosonOrchestrationHandler is
      * - When agent id is non zero:
      *   - If Agent does not exist
      *   - If the sum of agent fee amount and protocol fee amount is greater than the offer fee limit
+     * - _to is not the BosonVoucher contract address or the BosonVoucher contract owner
      *
      * @dev No reentrancy guard here since already implemented by called functions. If added here, they would clash.
      *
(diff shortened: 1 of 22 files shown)


contracts/protocol/clients/voucher/BosonVoucher.sol
```diff
@@ -56,12 +56,14 @@ contract BosonVoucherBase is IBosonVoucher, BeaconClientBase, OwnableUpgradeable
     // Tell if preminted voucher has already been _committed
     mapping(uint256 => bool) private _committed;
 
+    mapping(uint256 => address) private _preMintOwners;
```
zajck, line 59: Storing address per every `tokenId` does not makes much sense, since it invalidates all minting gas savings that were introduced with ranges. The main reasons to introduce ranges was to avoid storing the same address all over the place.
If we wanted to store address per token id it would be better to just scrap everything we have, and make a simple loop where we call existing OZ `_mint` method, which would be more efficient and less error prone (no need to store ranges, no need to do binary search).
I run limit estimation and now max preminted vouchers (in single tx) dropped to arround 1.000, while before it was > 10.000.

It would be better to just add `owner" field to existing `Range` struct.

contracts/protocol/clients/voucher/BosonVoucher.sol
```diff
@@ -217,42 +219,53 @@ contract BosonVoucherBase is IBosonVoucher, BeaconClientBase, OwnableUpgradeable
...
+     *
+     * Reverts if:
+     * - Offer id is not associated with a range
+     * - Amount to mint is more than remaining un-minted in range
+     * - Too many to mint in a single transaction, given current block gas limit
+     * - Offer already expired
+     * - Offer is voided
+     * - _to address is zero address
+     *
+     * @param _to - the address to mint the vouchers to
```
zajck, line 257: If `_to` can be any address, this gives `assistant` probably too great power - they could mint vouchers to anybody (including themselves), and whoever the first owner is, gets the paid price (directly from marketplace), while in protocol it is sellers account that needs to provide the funds.

Also, as long as `preMint` that transfers funds to assistant exists, the problem of who gets the funds isn't solved.

I believe we don't even need two methods (`preMint` and `preMintTo`). Just `preMint` should be enough, where destination address should be some trusted address (for example `treasury`, `address(this)` or address that can be set only by the admin).

contracts/protocol/clients/voucher/BosonVoucher.sol
```diff
@@ -785,15 +798,59 @@ contract BosonVoucherBase is IBosonVoucher, BeaconClientBase, OwnableUpgradeable
         uint256 _tokenId,
         uint256 _offerId
     ) internal {
-        require(_from == owner(), NO_SILENT_MINT_ALLOWED);
```
zajck, line 788: Removing this allows anyone (who has non zero balance) to simply transfer preminted vouchers, even if they are not owner of the preminted vouchers.
So if seller has two offers, one cheap and one expensive, attacker can just buy the cheaper, and then steals the expensive one without paying anything.
It must be checked that `_from` actually owns the `tokenId`

This is the test I used to validate that attack exists:
```javascript
            it.only("Transfer preminted voucher, but from is not the voucher owner", async function () {
              // make sure "rando" has non-zero balance. In reality, that would be easy to achieve by just buying something cheap.
              await bosonVoucher.connect(assistant)[selector](assistant.address, rando.address, tokenId, ...additionalArgs);

              // next token id. Make sure that assistant is the owner
              tokenId = Number(tokenId)+1
              let tokenOwner = await bosonVoucher.ownerOf(tokenId);
              assert.equal(tokenOwner, assistant.address, "Seller is not the owner");

              // Following call should fail, since rando is not the owner of preminted voucher 
              await expect(
                bosonVoucher.connect(rando)[selector](rando.address, rando.address, tokenId, ...additionalArgs)
              ).to.be.reverted;
            });
```
Put this test in the place where "Transfer preminted voucher, but from is not the contract owner" was before. It should fail, but it doesn't.

zajck (changes requested):

contracts/protocol/clients/voucher/BosonVoucher.sol
zajck: As we discussed today, we could have two methods:
1. `preMint` (which could maybe be renamed to `premintToOwner`) and
2. `preMintToContract` which just mints vouchers to `address(this)`.

contracts/protocol/clients/voucher/BosonVoucher.sol
anajuliabit: but then it would be necessary to create a `reserveRangeTo` method

contracts/protocol/clients/voucher/BosonVoucher.sol
anajuliabit: fixed

contracts/protocol/bases/OfferBase.sol
```diff
@@ -309,9 +315,13 @@ contract OfferBase is ProtocolBase, IBosonOfferEvents {
         ProtocolLib.ProtocolCounters storage pc = protocolCounters();
         uint256 _startId = pc.nextExchangeId;
 
-        // Call reserveRange on voucher
         IBosonVoucher bosonVoucher = IBosonVoucher(protocolLookups().cloneAddress[offer.sellerId]);
-        bosonVoucher.reserveRange(_offerId, _startId, _length);
+
+        // _to must be the contract address or the contract owner
+        require(_to == address(bosonVoucher) || _to == bosonVoucher.owner(), INVALID_TO_ADDRESS);
```
zajck, line 321:
```suggestion
        require(_to == address(bosonVoucher) || _to == msgSender(), INVALID_TO_ADDRESS);
```

`bosonVoucher.owner()` is the assistant, who at the same time is the one calling `reserveRange`, so this is more efficient.

Probably you can msgSender to memory variable, since it's also used in `RangeReserved` event.

contracts/protocol/clients/voucher/BosonVoucher.sol
```diff
@@ -61,7 +61,7 @@ contract BosonVoucherBase is IBosonVoucher, BeaconClientBase, OwnableUpgradeable
      * variables without shifting down storage in the inheritance chain.
      * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
      */
-    uint256[43] private __gap;
+    uint256[42] private __gap;
```
zajck, line 64: This should stay 43, since we don't have any new storage variable anymore.

contracts/protocol/clients/voucher/BosonVoucher.sol
```diff
@@ -240,19 +247,19 @@ contract BosonVoucherBase is IBosonVoucher, BeaconClientBase, OwnableUpgradeable
...
 
-        // Pre-mint the range to the seller
+        // Pre-mint the range
         uint256 tokenId;
-        address seller = owner();
         for (uint256 i = 0; i < _amount; i++) {
             tokenId = start + i;
-            emit Transfer(address(0), seller, tokenId);
+
+            emit Transfer(address(0), range.owner, tokenId);
```
zajck, line 255: Please store `range.owner` to memory variable, otherwise it's reading from the storage in every iteration.

contracts/protocol/facets/OrchestrationHandlerFacet1.sol
```diff
@@ -162,6 +164,7 @@ contract OrchestrationHandlerFacet1 is PausableBase, SellerBase, OfferBase, Grou
         OfferDurations calldata _offerDurations,
         uint256 _disputeResolverId,
         uint256 _reservedRangeLength,
+        address _to,
```
zajck, line 167: Interface `contracts/interfaces/handlers/IBosonOrchestrationHandler.sol` should also be updated to match implementation (same for all functions).

scripts/domain/Range.js
```diff
@@ -119,6 +120,15 @@ class Range {
     return bigNumberIsValid(this.lastBurnedTokenId, { optional: true });
   }
 
+  /**
+   * Is this Range instance's assistant field valid?
```
zajck, line 124:
```suggestion
   * Is this Range instance's owner field valid?
```

test/protocol/clients/BosonVoucherTest.js
```diff
@@ -407,6 +454,64 @@ describe("IBosonVoucher", function () {
...
+          let tokenOwner = await bosonVoucher.ownerOf(tokenId);
+          assert.equal(tokenOwner, bosonVoucher.address, `Wrong token owner for token ${tokenId}`);
+        }
+
+        // Token that is inside a range, but wasn't preminted yet should not have an owner
+        await expect(bosonVoucher.ownerOf(Number(amount) + Number(start) + 1)).to.be.revertedWith(
+          RevertReasons.ERC721_NON_EXISTENT
+        );
+
+        // Seller's balance should be updated for the total mint amount
```
zajck, line 505:
```suggestion
        // Contract's balance should be updated for the total mint amount
```

zajck (changes requested):

contracts/protocol/bases/OfferBase.sol
anajuliabit: @zajck but what if the assistant has transferred the voucher ownership?

contracts/protocol/clients/voucher/BosonVoucher.sol
anajuliabit: forgot to put it back, done

contracts/protocol/clients/voucher/BosonVoucher.sol
anajuliabit: done

scripts/domain/Range.js
anajuliabit: fixed

contracts/protocol/facets/OrchestrationHandlerFacet1.sol
anajuliabit: done

test/protocol/clients/BosonVoucherTest.js
anajuliabit: fixed

contracts/interfaces/handlers/IBosonOrchestrationHandler.sol
```diff
@@ -162,6 +162,7 @@ interface IBosonOrchestrationHandler is
      * - When agent id is non zero:
      *   - If Agent does not exist
      *   - If the sum of agent fee amount and protocol fee amount is greater than the offer fee limit
+     * - _to is not the BosonVoucher contract address or the BosonVoucher contract owner
```
mischat, line 165: this requires a space :)

zajck (approved):
LGTM

mischat (approved):
This looks good to me, thanks

PR #591, opened 2023-03-23
Prevent renounceOwnership

### LINKED ISSUE #588: Boson voucher owner should not be able to renounceOwnership
### Description:

`BosonVoucher` inherits `OwnableUpgradeable` which implements method `renounceOwnership()`. If owner calls it, it loses its owner privileges, therefore it can no longer call methods with `onlyOwner` modifier. Affected functions are `preMint`, `burnPremintedVouchers`, `setContractURI`, `callExternalContract` and `setApprovalForAllToContract`.

If that happens, seller is not totally broken, since updating assistant address in the protocol also sets it as a new contract owner. Therefore this is then merely an inconvenience that can happen. It is still better to prevent it. 

Also, since `transferOwnership` is restricted (owner cannot call it directly) it does not make sense to leave a similar function unrestricted. Most likely renouncing ownership should be part of acount deactivation inside the protocol.

### Recommendation:
Override `renounceOwnership()` in `BosonVoucher` and revert if it is called.

zajck:
Fix #588

### DIFF
--- a/contracts/protocol/clients/voucher/BosonVoucher.sol
+++ b/contracts/protocol/clients/voucher/BosonVoucher.sol
@@ -571,6 +571,15 @@ contract BosonVoucherBase is
     }
 
     /**
+     * @notice Overriding renounceOwnership() from OwnableUpgradeable, so it's not possible to renounce ownership.
+     *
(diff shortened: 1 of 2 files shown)


anajuliabit (approved):
LGTM

mischat (approved):
LGTM

Issue #590, opened 2023-03-23
Support multiple collections per seller


zajck:
Implement proposal from https://github.com/bosonprotocol/BPIPs/pull/14 which introduces multiple boson voucher contracts per one seller.

zajck:
Implemented in #592

PR #592, opened 2023-03-28
Multiple collections per seller

zajck:
PR implements #590.

- In `contracts/domain/BosonTypes.sol`:
  - added new type `Collection` which contains collection address and its external id
  - `collectionIndex` is added to `Offer` struct
- in `contracts/interfaces/handlers/IBosonAccountHandler.sol` and `contracts/protocol/facets/SellerHandlerFacet.sol` two methods are added:
  - `createNewCollection` that creates an additional collection for the seller
  - `getSellersCollections` which returns default boson voucher address and list of additional collections
- in `contracts/protocol/facets/ExchangeHandlerFacet.sol`, 
  - all methods that previously interacted with clone address now get the correct clone address, based on collectionIndex, stored as the part of the offer
- VoucherClone `initializeVoucher` is changed to accept `_collectionIndex` and add it as part of the name and symbol. Might still be changed: https://github.com/bosonprotocol/BPIPs/discussions/15#discussioncomment-5402214
- added a new event `CollectionCreated` which is emitted when a new collection is created
- added/updated domain script and tests for `Collection` and `Offer`
- added unit tests for use cases with collections


contracts/protocol/bases/ProtocolBase.sol
```diff
@@ -701,4 +701,15 @@ abstract contract ProtocolBase is PausableBase, ReentrancyGuardBase {
         // Determine existence
         exists = (_exchangeId > 0 && condition.method != EvaluationMethod.None);
     }
+
```
anajuliabit, line 704: could you please add Natspec comments here?

contracts/protocol/bases/SellerBase.sol
```diff
@@ -153,14 +153,16 @@ contract SellerBase is ProtocolBase, IBosonAccountEvents {
...
      * @param _assistant - address of the assistant
      * @param _voucherInitValues - the fully populated BosonTypes.VoucherInitValues struct
      * @return cloneAddress - the address of newly created clone
      */
     function cloneBosonVoucher(
         uint256 _sellerId,
+        uint256 _collectionIndex,
         address _assistant,
-        VoucherInitValues calldata _voucherInitValues
+        VoucherInitValues memory _voucherInitValues
```
anajuliabit, line 165: can you explain why did you change this?

test/protocol/ExchangeHandlerTest.js
```diff
@@ -1722,6 +1821,36 @@ describe("IBosonExchangeHandler", function () {
...
+        // expected address of the first additional collection
+        const additionalCollectionAddress = calculateContractAddress(accountHandler.address, "2");
+        const additionalCollection = await ethers.getContractAt("BosonVoucher", additionalCollectionAddress);
+
+        // Revoke the voucher, expecting event
+        await expect(exchangeHandler.connect(assistant).revokeVoucher(exchange.id))
+          .to.emit(additionalCollection, "Transfer")
+          .withArgs(buyer.address, ethers.constants.AddressZero, tokenId);
+      });
+
```
anajuliabit, line 1853: I suppose it would be more logical if we organized test contexts based on previous steps required instead of always grouping them by function. This way, we could avoid repeating a lot of pre-test setups. Sometimes I also feel that our test organization leads to test duplication. For example, if we want to test a code block in `commitToOfferInternal`, we end up repeating the same test at least three times because we have three functions that call `commitToOfferInternal` (commitToOffer, commitToPreMintedOffer, and more recently commitToConditionalOffer).

I'm not saying that you should do this in this PR, just wanted to bring this topic up for future discussions

anajuliabit (changes requested):
Looks really good.

Just some thoughts

test/protocol/ExchangeHandlerTest.js
zajck: I agree that some test files are a bit out of hand, and would benefit from some refactoring.

But I'm not sure that the different grouping actually generally solves the duplication, it just turns the code around (it requires that some other part is duplicated).

For example, consider these two organizations:
**opt 1:**
```
context A
    setup A
    ... other tests with setup A ...
    test D with an additional setup D

context B
   setup B
   ... other tests with setup B ...
   test D with an additional setup D
```
**opt 2:**
```
context D
    setup D
    test D with an additional setup A
    test D with an additional setup B

context XY
   setup XY
   ...
   test X with setup A
   test Y with setup B
```

In both cases the `test D` is the test that is a "duplicate", except that tested functions A and B are different (let's say `commitToOffer`, `commitToPreMintedOffer`). Test D requires a setup that is a combination of setups for A and D or B and D.

Opt 1 is similar to what we have now, so the top context is based on the function and we later duplicated the code for test D.
Now if we turn this around and group it by test D, we still need two tests, where one includes A specific setup and the other has B specific setup. We sure can do that. But there will still exist other contexts that will require either setup A or setup B. So instead of setup D being duplicated across contexts, now contexts A and B will be duplicated.

And in some cases, opt 2 is indeed better (for example, if setup D is long compared to setup A/B). So if we just wanted to optimize the tests we could regroup them. And that would lead to scenario-based contexts. We had that in v1 and I can tell you it was a nightmare to maintain. When some feature was added/removed it was sometimes even hard to accurately pinpoint into which scenario(s) we should add it.

Also, I think it makes sense to stick to the same option across all tests. And if I'm choosing between what have now and setup-based contexts, I'm sticking with the function based.

But I agree a lot of parts could be refactored to reuse the same code. Because now, if we make a change that affects a "duplicate" test, we need to go and find all affected instances. Ideally, the duplicated part would be a sort of util function (can be on a file level if other tests don't need it) and the test would simply invoke that.

anajuliabit: Utility functions make sense to me. We can start migrating to functions when we need to add or refactor a test. A separate file would be better IMO, as we could reuse it in different facets, such as ExchangeHandlerTest and OrchestrationHandlerTest. I believe this would help make the codebase more legible, as our test files are currently too large.

anajuliabit (dismissed):
LGTM

contracts/interfaces/handlers/IBosonAccountHandler.sol
```diff
@@ -353,6 +370,17 @@ interface IBosonAccountHandler is IBosonAccountEvents {
...
+    /**
+     * @notice Gets the details about a seller's collections.
+     *
+     * @param _sellerId - the id of the seller to check
+     * @return defaultVoucherAddress - the address of the default voucher contract for the seller
+     * @return additionalCollections - an array of additional collections that the seller has created
+     */
+    function getSellersCollections(
+        uint256 _sellerId
+    ) external view returns (address defaultVoucherAddress, BosonTypes.Collection[] memory additionalCollections);
```
mischat, line 382: Do we worry about the Collection[] being too big ? Do we need something like a `getCollection(CollectionID)` or something ?

mischat: even if this is a good idea, i wouldn't block on it ...

mischat (approved):
This looks good to me !

contracts/interfaces/handlers/IBosonAccountHandler.sol
anajuliabit: I also think is a good idea, nice catch. 
I create an issue so this doesn't get loss 
https://github.com/bosonprotocol/boson-protocol-contracts/issues/709

anajuliabit (approved):
LGTM

mischat:
@anajuliabit thanks for creating the issue.

PR #595, opened 2023-03-30
Refactor tests

zajck:
Closes #593 

I made three changes:
- added a new util function `setupTestEnvironment` which deploys protocol. It can be reused across most of the tests. There are some exceptions (`ConfigHandlerTest`, `ProtocolDiamondTest` and `ProtocolInitializationHandlerTest`) where tests were written in a way that using `setupTestEnvironment` did not make sense
- First `beforeEach` in most of the test files was changed into `before` block, where we take evm snapshot at the end of the deployment and then after each test call `evm_revert`. This significantly improved execution time. To take snapshot and revert back to it I added two more util functions (`getSnapshot`, `revertToSnapshot`)
- I changed algorithm in `split-unit-tests-into-chunks` to make more even chunks which should on average result in lower execution time.

Issue #596, opened 2023-03-31
Remove redundant check on createTwinInternal


anajuliabit:
The code block at [https://github.com/bosonprotocol/boson-protocol-contracts/blob/100b9b05599f12a77b93f9e7986c207df93e4ab9/contracts/protocol/bases/TwinBase.sol#L79-L93](https://github.com/bosonprotocol/boson-protocol-contracts/blob/100b9b05599f12a77b93f9e7986c207df93e4ab9/contracts/protocol/bases/TwinBase.sol#L79-L93) is redundant and should be removed. 

The check for overlapping twins is already performed on lines [https://github.com/bosonprotocol/boson-protocol-contracts/blob/100b9b05599f12a77b93f9e7986c207df93e4ab9/contracts/protocol/bases/TwinBase.sol#L100-L106](https://github.com/bosonprotocol/boson-protocol-contracts/blob/100b9b05599f12a77b93f9e7986c207df93e4ab9/contracts/protocol/bases/TwinBase.sol#L100-L106), so it is not necessary to perform it on lines 79-93.

The lookup `twinIdsByTokenAddressAndBySeller` should be deprecated, as it is only used for this code block.

mischat:
@anajuliabit, in your description above, you say that the lines 

Lines 79 to 93 in [100b9b0](https://github.com/bosonprotocol/boson-protocol-contracts/commit/100b9b05599f12a77b93f9e7986c207df93e4ab9)

could be removed, as the check happens on (the same) lines : 

Lines 79 to 93 in [100b9b0](https://github.com/bosonprotocol/boson-protocol-contracts/commit/100b9b05599f12a77b93f9e7986c207df93e4ab9). 

Could you please link to where the check is already happenings for me, as it will make it easier for me to review the resultant PR, thanks?

anajuliabit:
@mischat sorry, description updated

mischat:
thanks, this will make it easier for me to review! 👍

PR #597, opened 2023-04-03
Code coverage 2.2.0

zajck:
This PR already includes changes from #583, #589, #591 and #595 which should be reviewed first.

The first run with a new coverage version.

- Added missing tests
- Updated config
- Updated npm script
- Added coverage to CI


test/protocol/TwinHandlerTest.js
```diff
@@ -377,12 +377,10 @@ describe("IBosonTwinHandler", function () {
...
         it("The twins region of protocol is paused", async function () {
           // Pause the twins region of the protocol
-          await pauseHandler
-            .connect(pauser)
-            .pause([PausableRegion.Offers, PausableRegion.Twins, PausableRegion.Bundles]);
+          await pauseHandler.connect(pauser).pause([PausableRegion.Twins]);
 
           // Attempt to Remove a twin, expecting revert
-          await expect(twinHandler.connect(assistant).removeTwin(twin.id)).to.revertedWith(RevertReasons.REGION_PAUSED);
+          await expect(twinHandler.connect(assistant).createTwin(twin)).to.revertedWith(RevertReasons.REGION_PAUSED);
```
anajuliabit, line 383: Why did you remove the `removeTwin` test?

test/protocol/TwinHandlerTest.js
```diff
@@ -557,6 +555,7 @@ describe("IBosonTwinHandler", function () {
           await twinHandler.connect(assistant).createTwin(twin);
 
           // Create new twin with same token address
+          // twin.supplyAvailable = "2"
```
anajuliabit, line 558: 👀

anajuliabit (changes requested):

test/protocol/TwinHandlerTest.js
zajck: This line was wrong since it's inside `createTwin()` context. There were two tests for paused regions that used removeTwin and none that used `createTwin`.

Test for pausable region for `removeTwin` is in line 724.

test/protocol/TwinHandlerTest.js
zajck: I guess I was testing my findings that led to #596 and I forgot to uncomment this.
I think it shouldn't affect the result, so I uncommented it.

mischat (approved):
LGTM

anajuliabit (approved):

PR #598, opened 2023-04-12
Zora wrapper

zajck:
PoC for wrapper for uncompatibile PD mechanisms.
This works with Zora Auction House.

### DIFF
--- a/contracts/example/ZoraWrapper/ZoraWrapper.sol
+++ b/contracts/example/ZoraWrapper/ZoraWrapper.sol
@@ -10,6 +10,7 @@ import { ERC721 } from "./../support/ERC721.sol";
 import { IERC721Metadata } from "./../support/IERC721Metadata.sol";
 import { IERC721 } from "../../interfaces/IERC721.sol";
 import { IERC165 } from "../../interfaces/IERC165.sol";
+import "../../domain/BosonConstants.sol";
 
 /**
  * @title ZoraWrapper
@@ -26,6 +27,7 @@ contract ZoraWrapper is BosonTypes, Ownable, ERC721 {
     address private immutable voucherAddress;
     address private immutable zoraAuctionHouseAddress;
     address private immutable protocolAddress;
+    address private immutable wethAddress;
 
     uint256 private pendingTokenId;
     mapping(uint256 => uint256) public price;
@@ -40,20 +42,22 @@ contract ZoraWrapper is BosonTypes, Ownable, ERC721 {
     constructor(
         address _voucherAddress,
         address _zoraAuctionHouseAddress,
-        address _protocolAddress
+        address _protocolAddress,
+        address _wethAddress
     ) ERC721(getVoucherName(_voucherAddress), getVoucherSymbol(_voucherAddress)) {
         voucherAddress = _voucherAddress;
         zoraAuctionHouseAddress = _zoraAuctionHouseAddress;
         protocolAddress = _protocolAddress;
+        wethAddress = _wethAddress;
 
         // Approve Zora Auction House to transfer wrapped vouchers
-        setApprovalForAll(_zoraAuctionHouseAddress, true);
-        setApprovalForAll(msg.sender, true); // msg.sender is the owner of this contract and must be approved to transfer wrapped vouchers to Auction House
+        _setApprovalForAll(address(this),_zoraAuctionHouseAddress, true);
+        _setApprovalForAll(address(this), msg.sender, true); // msg.sender is the owner of this contract and must be approved to transfer wrapped vouchers to Auction House
     }
 
     function getVoucherName(address _voucherAddress) internal view returns (string memory) {
         string memory name = IERC721Metadata(_voucherAddress).name();
-        return string(abi.encodePacked("WRAPPED", name));
+        return string(abi.encodePacked("Wrapped ", name));
     }
 
     function getVoucherSymbol(address _voucherAddress) internal view returns (string memory) {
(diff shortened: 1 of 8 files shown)


contracts/example/ZoraWrapper/ZoraWrapper.sol
```diff
@@ -0,0 +1,241 @@
...
+     * @param _tokenId The token id.
+     */
+    function getCurrentBalance(uint256 _tokenId) internal returns (uint256) {
+        address exchangeToken = cachedExchangeToken[_tokenId];
+
+        // If exchange token is not known, get it from the protocol.
+        if (exchangeToken == address(0)) {
+            uint256 offerId = _tokenId >> 128; // OfferId is the first 128 bits of the token ID.
+
+            // Get Boson offer. Don't explicitly check if the offer exists, since existance of the token implies it does.
```
anajuliabit, line 204: > Don't explicitly check if the offer exists, since existence of the token implies it does

if `_tokenId` is a token id previous to `v2.2.0` ,`offerId` here will be 0 here and fetching `getOffer(0)` would return `[exist: false, offer: [...offer, exchangeToken: addressZero]` so we would assume exchangeToken as being `wethAddress` when actually we don't know if this is the correct exchangeToken as we fetch offer 0 and not true offer id. Does that make sense? maybe I'm missing something,

contracts/example/ZoraWrapper/ZoraWrapper.sol
```diff
@@ -0,0 +1,241 @@
...
+    /**
+     * @notice Wraps the voucher, transfer true voucher to itself and funds to the protocol.
+     *
+     * Reverts if:
+     *  - caller is not the contract owner
+     *
+     * @param _tokenId The token id.
+     */
+    function wrap(uint256 _tokenId) external onlyOwner {
+        // should wrapping be limited to onlyOwner?
```
anajuliabit, line 103: I think we should allow at least _tokenId owner to call, but for that we also need to change the logic in unwrap function not to check if caller is contract owner.

anajuliabit (changes requested):

contracts/example/ZoraWrapper/ZoraWrapper.sol
```diff
@@ -0,0 +1,245 @@
...
+        address _from,
+        address _to,
+        uint256 _tokenId
+    ) internal virtual override(ERC721) {
+        if (_from == zoraAuctionHouseAddress && _to != address(this)) {
+            // Auction is over, and wrapped voucher is being transferred to voucher owner
+            // If recipient is address(this), it means the auction was canceled and price updating can be skipped
+
+            // If some token price is not know yet, update it now
+            if (pendingTokenId != 0) updatePendingTokenPrice(pendingTokenId);
```
anajuliabit, line 172: there is no need to pass `pendingTokenId` as a parameter

anajuliabit (changes requested):

contracts/example/ZoraWrapper/ZoraWrapper.sol
```diff
@@ -0,0 +1,245 @@
...
+            price[pendingTokenId] = getCurrentBalance(_tokenId);
+            pendingTokenId = _tokenId;
+        }
+
+        super._beforeTokenTransfer(_from, _to, _tokenId);
+    }
+
+    function updatePendingTokenPrice(uint256 _tokenId) internal {
+        price[pendingTokenId] = getCurrentBalance(_tokenId) - price[pendingTokenId];
+    }
```
anajuliabit, line 184:
```suggestion
    function updatePendingTokenPrice() internal {
           uint256 tokenId = pendingTokenId;
           price[tokenId] = getCurrentBalance(tokenId) - price[tokenId];
    }
```

anajuliabit (changes requested):

contracts/example/ZoraWrapper/ZoraWrapper.sol
zajck: Removed.

contracts/example/ZoraWrapper/ZoraWrapper.sol
```diff
@@ -0,0 +1,245 @@
...
+        if (_from == zoraAuctionHouseAddress && _to != address(this)) {
+            // Auction is over, and wrapped voucher is being transferred to voucher owner
+            // If recipient is address(this), it means the auction was canceled and price updating can be skipped
+
+            // If some token price is not know yet, update it now
+            if (pendingTokenId != 0) updatePendingTokenPrice(pendingTokenId);
+
+            // Store current balance and set the pending token id
+            price[pendingTokenId] = getCurrentBalance(_tokenId);
+            pendingTokenId = _tokenId;
```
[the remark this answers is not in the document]
zajck: True. Or, logically equivalent is if I change `price[pendingTokenId]` to `price[_tokenId]` and have on SLOAD less.

contracts/example/ZoraWrapper/ZoraWrapper.sol
zajck: Fixed.

test/integration/price-discovery/auction.js
```diff
@@ -223,4 +223,129 @@ describe("[@skip-on-coverage] auctionProtocol integration", function() {
 
     expect(await bosonVoucher.ownerOf(tokenId)).to.equal(buyer.address);
   });
+
+  it.only("Works with Zora auction", async function () {
```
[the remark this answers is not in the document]
zajck: Removed.

anajuliabit (approved):
LGTM

Issue #603, opened 2023-04-14
Emit `VouchersPreminted` event


zajck:
Calling `preMint` does not emit a specific event. Just listening to `Transfer` events is not enough to distinguish between pre-minting and normal minting as part of ordinary `commitToOffer`.
On the other hand, listening to `RangeReserved` also does not give all the information about the preminted vouchers, since it's not necessary to premint the whole range at once.

Recommendation:

Add an event `vouchersPreminted(uint256 indexed offerId, uint256 startTokenId, uint256 endTokenId)`

zajck:
Fixed with #604

PR #604, opened 2023-04-14
Emit 'VouchersPreminted' event

anajuliabit:
Solve https://github.com/bosonprotocol/boson-protocol-contracts/issues/603

### DIFF
--- a/scripts/config/supported-interfaces.js
+++ b/scripts/config/supported-interfaces.js
@@ -28,6 +28,8 @@ const interfaceImplementers = {
   ProtocolInitializationHandlerFacet: "IBosonProtocolInitializationHandler",
 };
 
+const interfacesWithMultipleArtifacts = ["IERC2981", "IERC1155", "IERC165", "IERC721"];
+
 let interfacesCache; // if getInterfaceIds is called multiple times (e.g. during tests), calculate ids only once and store them to cache
 async function getInterfaceIds(useCache = true) {
   let interfaceIds = {};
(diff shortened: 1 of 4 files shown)


mischat (changes requested):
One of the unit tests seems to be failing. 

Also, out of interest, does adding an event change the interface, do we need to recalculate the "supportedInterface" ?

scripts/config/supported-interfaces.js
```diff
@@ -52,9 +54,15 @@ async function getInterfaceIds(useCache = true) {
   for (const iFace of interfaces) {
```
zajck, line 54: Since `IERC721` is in `interfacesWithMultipleArtifacts` and now uses its full name, you should also change that `skipBaseCheck[contracts/interfaces/IERC721.sol:IERC721]=false`  (not only `skipBaseCheck[IERC721]=false` as it is now).

zajck (changes requested):
I suppose that you needed to add handling of multiple artefacts since some clashed with some of your submodules.

Technically, this is not yet needed for this PR, but ok, probably will be useful in the future, so nice that you did it.

.gitmodules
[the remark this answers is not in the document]
anajuliabit: done

scripts/config/supported-interfaces.js
anajuliabit: fixed

mischat:
@anajuliabit looks like the unit tests are still failing : 

https://github.com/bosonprotocol/boson-protocol-contracts/actions/runs/4721809753/jobs/8375583900?pr=604

zajck (approved):

PR #606, opened 2023-04-21
Upgrade hooks

zajck:
Sometimes the upgrader needs to perform an action right before or after the upgrade.
For example, to upgrade to 2.2.0, the protocol must be paused, and the upgrader must get the next exchange id and add it to the configuration file. This can be done manually, but it's cumbersome, error-prone or sometimes it simply doesn't work (for example calling `pause` on louper gives an error).

This PR adds the possibility to write a pre- and post- upgrade scripts, which are executed when `upgrade-facets` script is run.


anajuliabit (approved):
Good job here!

mischat (approved):
LGTM

PR #614, opened 2023-04-24
Add metadataUri to seller struct

anajuliabit:
This PR also reverts when no field was updated on seller and DR updates to avoid expending unnecessary gas.


contracts/protocol/facets/DisputeResolverHandlerFacet.sol
```diff
@@ -289,6 +289,8 @@ contract DisputeResolverHandlerFacet is IBosonAccountEvents, ProtocolBase {
                 sender
             );
         }
+
+        require(updateApplied || needsApproval, NO_UPDATE_APPLIED);
```
mischat, line 293: Not sure i understand this line, but otherwise this WIP looks good. Why is this needed, and how come it has been added to the DisputeResolverHandler, is this just so that update methods throw in a consistent way ?

mischat (commented):
Otherwise this WIP looks good and should unblock the dapp / core components work

contracts/protocol/facets/DisputeResolverHandlerFacet.sol
anajuliabit: If no field has been updated or requested to be updated, there is no reason to execute the transaction. It is better to revert and return at least some gas to the user.

zajck (dismissed):
LGTM

contracts/protocol/facets/DisputeResolverHandlerFacet.sol
mischat: Ok, got it.

mischat (approved):
Thanks for this ! Nice and speedy turn around ... we should cut v2.2.1 now

zajck (dismissed):

zajck (approved):

PR #615, opened 2023-04-25
2.2.1 mumbai test addresses

anajuliabit:
- 2.2.1 mumbai test addresses
- Correctly handle constructor arguments


addresses/80001-mumbai-test.json
```diff
@@ -36,156 +45,159 @@
...
-      "interfaceId": "0x0d8e6e2c"
+      "interfaceId": "0xd9ea8317"
     },
     {
-      "name": "OrchestrationHandlerFacet2",
-      "address": "0x6286d3d8E5ceCB5F17f4255957A5926450d37C21",
-      "args": []
+      "name": "ExchangeHandlerFacet",
+      "address": "0x9AE3F5C157656C4e86b5F8340F399Ee33C62792e",
+      "args": [],
```
zajck, line 118: `args` should not be empty for facets with a constructor. That's important for contract verification.

Probably the deployment script needs to be changed to correctly store it.

zajck (changes requested):

addresses/80001-mumbai-test.json
mischat: what needs to happen here, @zajck ? Do we need to create an issue so that we don't forget, or do we want it to be fixed as a part of this PR ?

mischat (dismissed):
This looks good to me, i have left a comment in there...

addresses/80001-mumbai-test.json
zajck: The minimum we need to change is to manually update this file and update it with the value that was used during the deployment (1).

We can update the deploy script in another PR, but it must be added to `v2.2.1` release. 
And I believe that the existing `v2.2.1` release should be changed to `v2.2.1-rc.1` and changed to pre-release, since there are some details missing to be called a release.

anajuliabit: done

zajck (dismissed):

scripts/upgrade-facets.js
```diff
@@ -205,7 +205,13 @@ async function main(env, facetConfig) {
     contracts = contracts.filter((i) => i.name !== newFacet.name);
 
     const newFacetInterfaceId = interfaceIdFromFacetName(newFacet.name);
-    deploymentComplete(newFacet.name, newFacet.contract.address, [], newFacetInterfaceId, contracts);
+    deploymentComplete(
+      newFacet.name,
+      newFacet.contract.address,
+      newFacet.constructorArguments,
```
zajck, line 211:
```suggestion
      newFacet.constructorArgs,
```

Variable in `scripts/util/deploy-protocol-handler-facets.js` was renamed, so it must be here as well.

zajck, line 211:
```suggestion
      newFacet.constructorArgs,
```

Variable in `scripts/util/deploy-protocol-handler-facets.js` was renamed, so it must be here as well.

zajck (changes requested):

scripts/upgrade-facets.js
anajuliabit: fixed

mischat (approved):
This looks good to me

PR #624, opened 2023-05-09
Fix the bug in update seller

### LINKED ISSUE #623: Cannot update seller back to the same authTokenId & authTokenType
If you create a seller with an authTokenId X and authTokenType A, then you update it to authTokenId Y and authTokenType B and then back to authTokenId X and authTokenType A, you will get the error **_Auth token cannot be assigned to another entity of the same type_**. There seems to be a bug where the old authTokenId and/or authTokenType are not cleared.

zajck:
Fixes #623


anajuliabit (approved):
LGTM

mischat (approved):
This looks good to me, let's get it into v2.2.1-rc* and create a new RC. 

Thanks

PR #635, opened 2023-05-12
Fix token gating

### LINKED ISSUE #627: Protocol - Fix Token Gating
Currently, sellers can create token-gated offers with the following restrictions

1. Buyer must be holding a specific ERC721 (can be a SnapshotGate token) 
2. Buyer must be holding a minimum required balance of an ERC20/ERC1155/ERC721

and they can set the maximum number of commits that a certain address can make. This limit corresponds to the address, making the commits and is independent of tokenId. This means, that if a user holds a certain ERC721 token, needed to commit to an offer, they can commit and then transfer the token to another user, which can use the same token to commit to the offer.

If the seller wants to prevent this, they can:
1. create an offer which is gated with a specific ERC721 and limit it to 1 commit. But if they want to do this for a large number of tokens, they need to create a separate offer for each of them, which is not feasible (also, the must then search through the offers to find the one they can commit to).
2. create a snapshot gate, where they implement custom logic and they gate the offer with snapshot-gate tokens. This allows for greater flexibility but requires the deployment of a dedicated contract. Buyers then commit through this contract (not directly to the protocol). And since the snapshot represents only a point in time, users that obtain the ERC721 token later, cannot commit to the offer.

**Proposal**
Allow additional condition support directly in the protocol, where sellers can gate an offer with a range of ERC721 tokens, but each token can be used to commit only `n` times.

This would require introducing new lookup variables in the protocol and creating a new `commitToOffer`, which will accept an additional argument, where the buyer can specify the arbitrary data. In this case, the data will be a `tokenId` that buyer wants to use to commit.

anajuliabit:
Fix #627


contracts/protocol/libs/ProtocolLib.sol
```diff
@@ -182,6 +182,8 @@ library ProtocolLib {
         mapping(uint256 => BosonTypes.AuthToken) pendingAuthTokenUpdatesBySeller;
         // dispute resolver id => DisputeResolver
         mapping(uint256 => BosonTypes.DisputeResolver) pendingAddressUpdatesByDisputeResolver;
+        // groupId => tokenId => commit count (count how many times a token has been used as gate for this group)
+        mapping(uint256 => mapping(uint256 => uint256)) conditionalCommitsByTokenId;
```
zajck, line 186: Not that it matters that much, but `conditionalCommitsByAddress` has just different mapping order
`address => groupId => commit count`, (so groupId comes after the address), while in `conditionalCommitsByTokenId` it comes before `tokenId`.

Performance-wise it think there is no difference, and probably your choice is a bit more natural, but I think that we should be consistent here and apply the same order in both cases. And since by address lookup is already there, the new one should be changed.

contracts/interfaces/handlers/IBosonExchangeHandler.sol
```diff
@@ -43,6 +43,8 @@ interface IBosonExchangeHandler is IBosonExchangeEvents, IBosonFundsLibEvents, I
      */
     function commitToOffer(address payable _buyer, uint256 _offerId) external payable;
 
+    function commitToConditionalOffer(address payable _buyer, uint256 _offerId, uint256 _tokenId) external payable;
```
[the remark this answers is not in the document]
anajuliabit: added

contracts/protocol/bases/ProtocolBase.sol
```diff
@@ -519,13 +519,11 @@ abstract contract ProtocolBase is PausableBase, ReentrancyGuardBase {
      * Reverts if:
      * - Offer does not exist
      * - Offer already voided
-     * - Caller is not the seller
      *
      *  @param _offerId - the id of the offer to check
      */
     function getValidOffer(uint256 _offerId) internal view returns (Offer storage offer) {
```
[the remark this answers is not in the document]
anajuliabit: done

contracts/protocol/bases/GroupBase.sol
```diff
@@ -94,15 +100,22 @@ contract GroupBase is ProtocolBase, IBosonGroupEvents {
...
     /**
      * @notice Validates that condition parameters make sense.
      *
-     * Reverts if:
-     * - EvaluationMethod.None and has fields different from 0
-     * - EvaluationMethod.Threshold and token address or maxCommits is zero
-     * - EvaluationMethod.SpecificToken and token address or maxCommits is zero
+     * A invalid condition is one that fits any of the following criteria:
+     * - EvaluationMethod.None: any fields different from 0
+     * - EvaluationMethod.Threshold: token address, maxCommits or threshold is zero and length is not zero
```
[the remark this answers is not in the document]
anajuliabit: done

contracts/protocol/bases/GroupBase.sol
```diff
@@ -113,11 +126,31 @@ contract GroupBase is ProtocolBase, IBosonGroupEvents {
...
+            if (_condition.tokenType == TokenType.NonFungibleToken) {
+                valid = valid && _condition.threshold == 0;
+
+                if (_condition.tokenId != 0) {
+                    // SpecificToken with NonFungibleToken and tokenId should have length
+                    valid = valid && _condition.length > 0;
+                }
+            } else {
+                // SpecificToken with MultiToken should have threshold
+                valid = valid && _condition.threshold > 0;
```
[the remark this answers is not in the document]
anajuliabit: good catch, fixed

contracts/protocol/facets/ExchangeHandlerFacet.sol
```diff
@@ -79,17 +78,75 @@ contract ExchangeHandlerFacet is IBosonExchangeHandler, BuyerBase, DisputeBase {
...
+     * - Conditional commit requirements not met or already used
+     * - Offer price is in native token and caller does not send enough
+     * - Offer price is in some ERC20 token and caller also sends native currency
+     * - Contract at token address does not support ERC20 function transferFrom
+     * - Calling transferFrom on token fails for some reason (e.g. protocol is not approved to transfer)
+     * - Received ERC20 token amount differs from the expected value
+     * - Seller has less funds available than sellerDeposit
+     *
+     * @param _buyer - the buyer's address (caller can commit on behalf of a buyer)
+     * @param _offerId - the id of the offer to commit to
```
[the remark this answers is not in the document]
anajuliabit: added

contracts/protocol/facets/ExchangeHandlerFacet.sol
```diff
@@ -79,17 +78,75 @@ contract ExchangeHandlerFacet is IBosonExchangeHandler, BuyerBase, DisputeBase {
...
+        // For there to be a condition, there must be a group.
+        (bool exists, uint256 groupId) = getGroupIdByOffer(offer.id);
+
+        // Make sure the group exists
+        require(exists, NO_SUCH_GROUP);
+
+        // Get the condition
+        Condition storage condition = fetchCondition(groupId);
+
+        bool allow = authorizeCommit(_buyer, condition, groupId, _tokenId);
```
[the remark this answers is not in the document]
anajuliabit: Now only works with `commitToOffer`

contracts/protocol/bases/ProtocolBase.sol
```diff
@@ -535,6 +533,23 @@ abstract contract ProtocolBase is PausableBase, ReentrancyGuardBase {
...
+    /**
+     * @notice Gets offer and seller from protocol storage
+     *
+     * Reverts if:
+     * - Offer does not exist
+     * - Offer already voided
+     * - Seller assistant is not the caller
+     *
+     *  @param _offerId - the id of the offer to check
+     */
```
zajck, line 547: Just for consistency, maybe add return parameters to natspec.

And now looking at the places where it's used. apparently, it doesn't even have to return the seller.
I guess you can remove it from return. 
Maybe it then needs only a better name (I know I suggested this one :grin: ), so it doesn't sound like it's returning the seller. Sthi like, `getValidOfferWithSellerCheck`? I'm so bad at this, if you think of anything better, just use it instead.

zajck (changes requested):
Looks good, just another small request.

contracts/protocol/bases/ProtocolBase.sol
anajuliabit: done

.github/workflows/ci.yaml
```diff
@@ -134,7 +134,7 @@ jobs:
       - name: Check Code Coverage
         shell: bash
         run: |
-          MAX_SKIPPED=1 # at this moment msgData() in BosonVoucher is not covered
+          MAX_SKIPPED=2 # at this moment msgData() in BosonVoucher is not covered
```
zajck, line 137: @anajuliabit What is the 2nd statement that we allow to skip?

zajck (dismissed):
LGTM

.github/workflows/ci.yaml
anajuliabit: this was a mistake, thanks

zajck (dismissed):

contracts/protocol/facets/ExchangeHandlerFacet.sol
```diff
@@ -995,9 +995,6 @@ contract ExchangeHandlerFacet is IBosonExchangeHandler, BuyerBase, DisputeBase {
                 // Increment number of commits to the group for this address if they are allowed to commit
                 lookups.conditionalCommitsByAddress[_buyer][_groupId] = ++commitCount;
             }
-        } else {
-            // No condition set, so allow the commit
-            allow = true;
         }
```
zajck, line 998: Now that you removed this line, preminted offers for groups with the condition with EvaluationMethod `None` are not possible anymore.

Maybe it looks like an edge case, but I think it's a valid one and we should support preminted offers with None condition.

For example, a seller could create a preminted offer with a group that has a threshold condition. But later they decide to remove that condition and they can do it by calling `setGroupCondition` and setting it to None. We want that preminted offers still work after that.

zajck (changes requested):

contracts/protocol/facets/ExchangeHandlerFacet.sol
anajuliabit: actually this is not possible since would revert on line 200
https://github.com/bosonprotocol/boson-protocol-contracts/pull/635/files#diff-9734ea571f47efbe33f29fc8ab6a117aa407a1a4b8bdbb0a40771e9b0afad824R200

zajck: You added that "require" after my review. But that does not make it any better.

I'm saying that `commitToPreMintedOffer` should succeed if `condition.method == EvaluationMethod.None` and now it doesn't.

Also now preminted offers cannot be gated with `Multitoken` which again seems a very arbitrary restriction. 
I understand that preminted offers cannot work with conditions where buyers must communicate which `tokenId` is being used to commit, but all other cases should be supported.

contracts/protocol/bases/GroupBase.sol
```diff
@@ -126,7 +126,8 @@ contract GroupBase is ProtocolBase, IBosonGroupEvents {
             valid = (_condition.tokenAddress != address(0) &&
                 _condition.maxCommits > 0 &&
                 _condition.threshold > 0 &&
-                _condition.length == 0);
+                _condition.length == 0 &&
+                _condition.tokenId == 0);
```
zajck, line 130: If token is a multitoken, tokenId 0 is a valid one, so we should not prevent the creation of it.

contracts/protocol/facets/ExchangeHandlerFacet.sol
```diff
@@ -977,19 +975,26 @@ contract ExchangeHandlerFacet is IBosonExchangeHandler, BuyerBase, DisputeBase {
...
+            require(commitCount < _condition.maxCommits, MAX_COMMITS_TOKEN_REACHED);
+
             allow = holdsSpecificToken(_buyer, _condition, _tokenId);
 
             if (allow) {
                 // Increment number of commits to the group for this token id if they are allowed to commit
                 lookups.conditionalCommitsByTokenId[_tokenId][_groupId] = ++commitCount;
             }
         } else if (_condition.method == EvaluationMethod.Threshold) {
+            require(_condition.tokenType == TokenType.MultiToken ? _tokenId > 0 : _tokenId == 0, INVALID_TOKEN_ID);
```
zajck, line 990: This is not backwards compatible - there might already exist conditions with `condition.tokenType == TokenType.MultiToken` and `_tokenId == 0`.

contracts/protocol/facets/ExchangeHandlerFacet.sol
```diff
@@ -196,7 +196,10 @@ contract ExchangeHandlerFacet is IBosonExchangeHandler, BuyerBase, DisputeBase {
             Condition storage condition = fetchCondition(groupId);
 
             // Make sure condition is not SpecificToken as it is not supported for preminted offers
-            require(condition.method != EvaluationMethod.SpecificToken, CANNOT_COMMIT);
+            require(
+                condition.method == EvaluationMethod.Threshold && condition.tokenType != TokenType.MultiToken,
```
zajck, line 200: Why is `MultiToken` not allowed for preminted offers?

zajck (changes requested):

contracts/protocol/bases/GroupBase.sol
anajuliabit: I thought this was the decision we discussed in Slack yesterday. As @mischat said, the ERC1155 per-address condition flow should look like ERC721 and ERC20 per-address condition flow. (any token is accepted)

contracts/protocol/facets/ExchangeHandlerFacet.sol
anajuliabit: because caller must provide a additional `tokenId` parameter

contracts/protocol/bases/GroupBase.sol
anajuliabit: IMO it makes sense since is "per-address" not "per-token"

contracts/protocol/facets/ExchangeHandlerFacet.sol
anajuliabit: > I'm saying that commitToPreMintedOffer should succeed if condition.method == EvaluationMethod.None and now it doesn't.

You're right sorry

contracts/protocol/bases/GroupBase.sol
zajck: I actually wrote my comment wrong, since this condition is enforcing tokenId to be 0, it's not preventing it.

Anyway, just accepting any token is way too vague. I as a seller might want to allow commits only if the user has a certain amount of ERC1155 tokens, and not just any token.

For example, let's say I have an ERC1155 for a fantasy game. Let's say my tokenIds have the following representations:
0: active user token. Anyone that plays it, has one.
1: common sword
2: rare sword
3: epic sword
4: common armour
5: rare armour

Let's say that I want to offer something only to the users that have at least 1 "epic sword". I'd like to put tokenId = 3 in, but I can't. So if I make a offer, anyone, even if they have only the basic activity token, can commit.
And I cannot go about it, by making a condition with `EvaluationMethod.SpecificToken` and setting it to 3, because I cannot really handle maxCommits there. If I put 1, only 1 person can get it. If I put more (let's say 3), one user can just buy all 3 items. 

The thing that we discussed yesterday, I understood as adding additional support for ranges, but in a way that would not affect the existing conditions. For example, in 2.2.1, if the seller wants to offer something to users that have any type of sword, they need to create 3 offers (one for common, one for rare one for epic). Now with the new support for ranges, a seller could set `tokenId = 1` and `length = 3`, so any user that has one of these three tokens could commit to the offer.

Basically this means, that any combination of `tokenId` and `length` should be ok (and by "any" I mean a combination such that `tokenId+length < maxuint256`

contracts/protocol/bases/GroupBase.sol
```diff
@@ -113,11 +120,32 @@ contract GroupBase is ProtocolBase, IBosonGroupEvents {
...
+        } else {
+            // SpecificToken
+            valid = (_condition.tokenAddress != address(0) &&
+                _condition.maxCommits > 0 &&
+                _condition.tokenType != TokenType.FungibleToken); // FungibleToken not allowed for SpecificToken
+
+            if (_condition.tokenId != 0) {
+                // SpecificToken with tokenId should have length
+                valid = valid && _condition.length > 0;
```
zajck, line 139: We should also make sure that `length + tokenId` does not overflow, otherwise we can run into problems in `authorizeCommit`

contracts/protocol/facets/ExchangeHandlerFacet.sol
zajck: Referring to my other comment (in `validateCondition`), I think we should allow `length` and `tokenId` to be non-zero for `EvaluationMethod.Threshold`.

In that case, we could support both cases:
- when a range with length > 1, the user must specify which `tokenId` they want to use
- if the range has length 1, then it's unambiguous which `tokenId` is relevant and we can support preminted offers with multitoken condition.

contracts/protocol/facets/ExchangeHandlerFacet.sol
```diff
@@ -871,83 +951,105 @@ contract ExchangeHandlerFacet is IBosonExchangeHandler, BuyerBase, DisputeBase {
...
+        uint256 _tokenId
+    ) internal returns (bool allow) {
         // Cache protocol lookups for reference
         ProtocolLib.ProtocolLookups storage lookups = protocolLookups();
 
-        // Allow by default
-        bool allow = true;
+        if (_condition.method == EvaluationMethod.SpecificToken) {
+            // If condition has a token id, check that the token id is in range, otherwise accept any token id
+            if (_condition.tokenId > 0) {
```
zajck, line 971: I believe we should make this check based on `_condition.length > 0`.

It's a completely valid case that someone puts `_condition.tokenId = 0` and `_condition.length = 10` if they want to limit the range to the first 10 ids. But in this case, the range check would be skipped and someone with a token without

On the other hand, `_condition.length > 0` is a good indicator if the range needs to be validated or not.

contracts/protocol/bases/GroupBase.sol
anajuliabit: overflow check added

contracts/protocol/facets/ExchangeHandlerFacet.sol
anajuliabit: I used _condition. length > 1, because if length == 1 then we check _condition.tokenId and not _tokenId, which allow more use cases for conditional preminted vouchers

test/protocol/ExchangeHandlerTest.js
```diff
@@ -872,6 +915,143 @@ describe("IBosonExchangeHandler", function () {
         ).to.emit(exchangeHandler, "BuyerCommitted");
       });
 
+      context("Offer is part a group", async function () {
```
zajck, line 918:
```suggestion
      context("Offer is part of a group", async function () {
```

test/protocol/ExchangeHandlerTest.js
```diff
@@ -872,6 +915,143 @@ describe("IBosonExchangeHandler", function () {
...
+
+          await foreign721.connect(buyer).mint("123", 1);
+
+          await expect(bosonVoucher.connect(assistant).transferFrom(assistant.address, buyer.address, tokenId)).to.emit(
+            exchangeHandler,
+            "BuyerCommitted"
+          );
+        });
+
+        it("Offer is part of a group that has a per-wallet ERC721 condition", async function () {
```
zajck, line 1024: This test looks the same to me:
https://github.com/bosonprotocol/boson-protocol-contracts/blob/0e0f4a53311e4f28425e7bbac5a75c6d4bb825a8/test/protocol/ExchangeHandlerTest.js#L948-L971

test/protocol/ExchangeHandlerTest.js
```diff
@@ -982,10 +1162,119 @@ describe("IBosonExchangeHandler", function () {
...
+          groupId = "1";
+          offerIds = [offerId];
+
+          condition = mockCondition({
+            tokenAddress: foreign721.address,
+            threshold: "0",
+            maxCommits: "3",
+            tokenType: TokenType.NonFungibleToken, // ERC721
+            tokenId: "0",
+            method: EvaluationMethod.SpecificToken, // per-wallet
```
zajck, line 1258:
```suggestion
            method: EvaluationMethod.SpecificToken, // per-token
```

contracts/protocol/facets/ExchangeHandlerFacet.sol
```diff
@@ -80,17 +80,91 @@ contract ExchangeHandlerFacet is IBosonExchangeHandler, BuyerBase, DisputeBase {
...
+
+        // Make sure the group exists
+        require(exists, NO_SUCH_GROUP);
+
+        // Get the condition
+        Condition storage condition = fetchCondition(groupId);
+
+        require(condition.method != EvaluationMethod.None, GROUP_HAS_NO_CONDITION);
+
+        if (condition.length > 1) {
```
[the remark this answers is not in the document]
anajuliabit: good catch here, changed to >=

test/protocol/ExchangeHandlerTest.js
anajuliabit: This test was written incorrectly. It was supposed to say "Offer is part of a group that enforces per-wallet condition and range length == 1"

contracts/protocol/bases/GroupBase.sol
```diff
@@ -113,11 +125,53 @@ contract GroupBase is ProtocolBase, IBosonGroupEvents {
...
-            valid = (_condition.tokenAddress != address(0) && _condition.threshold == 0 && _condition.maxCommits > 0);
+                _condition.maxCommits == 0 &&
+                _condition.length == 0);
+        } else {
+            if (_condition.tokenId != 0) {
+                if (_condition.length == 0) {
+                    return false;
+                }
+
+                // Create local copy so we can use assembly to check for overflow
```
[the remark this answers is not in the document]
zajck: yes, you are right.

contracts/protocol/bases/GroupBase.sol
```diff
@@ -120,7 +121,8 @@ contract GroupBase is ProtocolBase, IBosonGroupEvents {
      * @return valid - validity of condition
      *
      */
-    function validateCondition(Condition memory _condition) internal pure returns (bool valid) {
+    function validateCondition(Condition memory _condition) internal pure returns (bool) {
```
zajck, line 124:
```suggestion
    function validateCondition(Condition calldata _condition) internal pure returns (bool) {
```
I just noticed this.
My last request, I promise :blush:

zajck (changes requested):
Looks good... But I managed to find one more little thing (that's been there for ages actually, but seems a good time to fix it).

contracts/protocol/bases/GroupBase.sol
anajuliabit: done :)

zajck (approved):
LGTM :+1:

test/protocol/ExchangeHandlerTest.js
```diff
@@ -724,7 +763,7 @@ describe("IBosonExchangeHandler", function () {
           ).to.revertedWith(RevertReasons.REGION_PAUSED);
         });
 
-        it("buyer address is the zero address", async function () {
+        it("buyer.address is the zero address", async function () {
```
zajck, line 766: Funny change, but I can live with it :)

zajck (approved):
Looks good :+1:

contracts/protocol/bases/GroupBase.sol
```diff
@@ -94,31 +94,74 @@ contract GroupBase is ProtocolBase, IBosonGroupEvents {
...
     }
 
     /**
      * @notice Validates that condition parameters make sense.
      *
-     * Reverts if:
-     * - EvaluationMethod.None and has fields different from 0
-     * - EvaluationMethod.Threshold and token address or maxCommits is zero
-     * - EvaluationMethod.SpecificToken and token address or maxCommits is zero
+     * A invalid condition is one that fits any of the following criteria:
```
mischat, line 103: s/A invalid/An invalid/ ... one for the future

contracts/protocol/bases/GroupBase.sol
```diff
@@ -94,31 +94,74 @@ contract GroupBase is ProtocolBase, IBosonGroupEvents {
...
+     * - EvaluationMethod.None: any field different from zero
+     * - EvaluationMethod.Threshold: 
+          -Token address, maxCommits, or threshold is zero.
```
mischat, line 106: s/-T/- T/ for later

mischat (approved):
looks good to me.

contracts/protocol/bases/GroupBase.sol
anajuliabit: I fixed on https://github.com/bosonprotocol/boson-protocol-contracts/pull/656 :)

contracts/protocol/bases/GroupBase.sol
anajuliabit: I fixed on https://github.com/bosonprotocol/boson-protocol-contracts/pull/656 :)

PR #638, opened 2023-05-16
Deprecate clerk

### LINKED ISSUE #629: Deprecate clerk role
Seller and dispute resolver entities have a "clerk" role.
The only role clerk has is to initiate the withdrawal of the entity's funds from the protocol to the entity's treasury address.

This responsibility can be moved to the assistant role, which already performs most of the protocol actions.

** To Do **
- [x] Depreacte "clerk" in the `Seller` and `DisputeResolver` entities. Although it won't be used anymore, it should remain in the struct to preserve the storage layout.
- [x] Change `withdrawFunds` method to allow the assistant to withdraw the funds to the treasury.

zajck:
Closes #629


test/example/ZoraWrapperTest.js
anajuliabit, line 1: I suppose this file shouldn't belong to this PR

anajuliabit (changes requested):
Could you also please update docs/task.md?

anajuliabit (approved):

contracts/protocol/facets/OrchestrationHandlerFacet1.sol
```diff
@@ -982,8 +987,9 @@ contract OrchestrationHandlerFacet1 is PausableBase, SellerBase, OfferBase, Grou
      * - The bundles region of protocol is paused
      * - The exchanges region of protocol is paused
      * - The orchestration region of protocol is paused
-     * - Caller is not the supplied assistant and clerk
+     * - Caller is not the supplied assistant and
```
mischat, line 990:
```suggestion
     * - Caller is not the supplied assistant
```

mischat (approved):
this LGTM ... i have suggested a minor copy change, otherwise, it makes sense, and nice to see no mention of the clerk in the contract, par the deprecated storage slots!

mischat (changes requested):
It looks like one of the unit tests isn't passing right now.

mischat:
@zajck it looks like we have unit tests failing, after merging in main

contracts/protocol/facets/OrchestrationHandlerFacet1.sol
zajck: Removed.

mischat (approved):
LGTM

mischat:
@anajuliabit ^^ you need to re-review, sorry

anajuliabit (approved):

PR #639, opened 2023-05-19
2.2.0 fix metaTransaction initialization

zajck:
`MetaTransactionsHandlerFacet` initialization required ADMIN role since it calls `setAllowlistedFunctions`.

This PR changes it, so the functions can be allowlisted during the initialization even without the ADMIN role.


mischat (approved):
This looks good to me!

Issue #636, opened 2023-05-15
Improve coverage CI report


anajuliabit:
At present, we are utilizing `solidity-coverage` for coverage assessment. There are, however, other coverage tools like https://github.com/coverallsapp/github-action or https://github.com/allure-framework that we could integrate into our CI process for better results. 

https://github.com/neonevm/neon-compatibility/- this project has successfully incorporated allure using hardhat  
https://github.com/compound-finance/comet - this one has integrated with coveralls.

zajck:
Done in #641

PR #641, opened 2023-05-23
Improve coverage CI report with coveralls

anajuliabit:
#636

This PR adds coveralls to GHA


albertfolch-redeemeum (approved):
LGTM

.github/workflows/ci.yaml
```diff
@@ -141,8 +139,38 @@ jobs:
...
+        with:
+          node-version: 16.14.x
+          cache: "npm"
+
+      - name: npm install
+        run: npm install
+
+      - name: Download code coverage results
+        uses: actions/download-artifact@v2
+        with:
```
zajck, line 163: image

Can you set node version to 16 to get rd of this warning?

package.json
```diff
@@ -64,8 +64,8 @@
     "manage-roles:polygon:mainnet": "npx hardhat manage-roles --network polygon --env prod >> logs/polygon.manage.roles.txt",
     "manage-roles:local": "npx hardhat manage-roles --network localhost >> logs/localhost.manage.roles.txt",
     "manage-roles:test": "npx hardhat manage-roles --network test --env test >> logs/test.manage.roles.txt",
-    "natspec-interface-id": "npx hardhat compile && node scripts/util/natspec-interfaceid.js",
```
zajck, line 67: Maybe I'm missing something, but it seems that `natspec-interfaceid.js` does not compile the contracts and just takes whatever is in the artefacts (which might be different from the latest contracts).
In GHA that's not an issue (since they are compiled just before), but locally we can't assume that user will call compile before calling this script.

zajck (changes requested):

.github/workflows/ci.yaml
anajuliabit: done

package.json
anajuliabit: you are right. At first, I thought we were compiling in both CI and pre-commit hook before we called `natspec-interface-id`, so I made this change. But we don't actually compile in pre-commit hook. And also, I can't assume everyone is using the hook for fixing interface ids as I do 😅

zajck (approved):
LGTM :+1:

mischat:
@anajuliabit i think we need to resolve a couple of conflicts here, and then we can get this is, given that it is the next one on the list.

zajck (dismissed):

zajck (approved):

mischat (approved):
LGTM

PR #644, opened 2023-05-24
Unpause individual regions

### LINKED ISSUE #631: Unpause individual regions
Currently, `unpause` unpauses all pausable regions at once.

There might exist some cases, where we'd like to unpause only some paused regions.

Make `unpause` similar to `pause`, so the caller can specify which regions should be unpaused.

anajuliabit:
Fix #631


zajck (changes requested):
I added some suggestions, most of them are using bitwise operations instead of common math operations.

I suggested this already to Clff when he implemented it, but he felt it was easier to understand it that way. 
I prefer bitwise operations since they are not only more efficient but allow us to get rid of some checks. But it's up to you to decide which way you will go. Since pausing/unpausing is not expected to be called often, it doesn't matter that much.

What I commented directly in the code is somewhat half-way toward what `togglePause` could be. Suggested changes actually allow to get rid of 
```
require(used[enumVal] != region, REGION_DUPLICATED);
```
and we can make the loop a bit different. So the final togglePause could be something like

```
function togglePause(BosonTypes.PausableRegion[] calldata _regions, bool _pause) internal {
        // Cache protocol status for reference
        ProtocolLib.ProtocolStatus storage status = protocolStatus();

        // Toggle all regions if none are specified.
        if (_regions.length == 0) {
            // Store the toggle scenario
            status.pauseScenario = _pause ? ALL_REGIONS_MASK : 0;
            return;
        }

        // Build the toggle scenario by summing the supplied
        // enum values, first converted to powers of two
        uint256 region;
        uint256 incomingScenario;

        // Calculate the incoming scenario as the summ of individual regions
        // Use "or" to get the correct value even if the same region is specified more than once
        for (uint256 i = 0; i < _regions.length; i++) {
            // Get enum value as power of 2
            region = 1 << uint256(_regions[i]);
            incomingScenario |= region;
        }

        // Store the toggle scenario
        if (_pause) {
            // for pausing, just "or" the incoming scenario with the existing one
            // equivalent to summation
            status.pauseScenario |= incomingScenario;
        } else {
            // for unpausing, "and" the inverse of the incoming scenario with the existing one
            // equivalent to subtraction
            status.pauseScenario &= ~incomingScenario;
        }
    }
```
Maybe this is premature optimization, maybe it's not. It just felt natural to me, so I wrote it here.
It's fine if you don't use it :relaxed:

anajuliabit:
@zajck, your suggestions were very valuable. I applied all of them.

zajck (approved):
LGTM :+1:

contracts/interfaces/handlers/IBosonPauseHandler.sol
```diff
@@ -38,4 +38,11 @@ interface IBosonPauseHandler is IBosonPauseEvents {
      * @param _regions - an array of regions to pause. See: {BosonTypes.PausableRegion}
      */
     function unpause(BosonTypes.PausableRegion[] calldata _regions) external;
+
+    /**
+     * @notice Returns the regions paused
+     *
+     * @return an array of regions that are currently paused. See: {BosonTypes.PausableRegion}
+     */
+    function getPausedRegions() external view returns (BosonTypes.PausableRegion[] memory);
```
zajck, line 47: I think it would be nice to have a named return value. We do it in most parts of the protocol.

contracts/protocol/facets/PauseHandlerFacet.sol
```diff
@@ -64,6 +64,46 @@ contract PauseHandlerFacet is ProtocolBase, IBosonPauseHandler {
...
+            }
+        }
+
+        // Create a new array with the correct size
+        BosonTypes.PausableRegion[] memory pausedRegions = new BosonTypes.PausableRegion[](count);
+        for (uint256 i = 0; i < count; i++) {
+            pausedRegions[i] = regions[i];
+        }
+
+        return pausedRegions;
```
zajck, line 104: There is a trick to adjust the length of a memory array, so you don't have to loop. Here it's not really needed, since these arrays will always be short. But it's worth remembering it if you ever need to do it on some longer arrays.

All you need to do is to change the first slot of the array which contains the array length, which can be done with
```
        // setting the correct number of regions
        assembly {
            mstore(regions, count)
        }

        return regions;
```

test/protocol/PauseHandlerTest.js
```diff
@@ -63,10 +65,18 @@ describe("IBosonPauseHandler", function () {
...
-    function regionsToScenario(regions) {
-      return regions.reduce((acc, region) => {
-        return acc + Math.pow(2, region);
-      }, 0);
+    function scenarioToRegions(scenario) {
+      const regions = [];
+      let region = 0;
+      while (scenario > 1) {
+        if (scenario % 2 === 1) {
+          console.log(region, scenario);
```
zajck, line 73: I guess `console.log` is not needed.

test/protocol/PauseHandlerTest.js
```diff
@@ -106,16 +120,14 @@ describe("IBosonPauseHandler", function () {
...
-
         // Pause the protocol, testing for the events
         await expect(pauseHandler.connect(pauser).pause(regions))
           .to.emit(pauseHandler, "ProtocolPaused")
           .withArgs(regions, pauser.address);
 
         // Check that both old and news regions are pause
-        const pauseScenario = await getStorageAt(pauseHandler.address, protocolStatusStorageSlotNumber);
-        expect(BigNumber.from(pauseScenario).toNumber(), "Protocol not paused").to.equal(expectedScenario);
+        const pausedRegions = await pauseHandler.getPauseStatus();
```
zajck, line 129:
```suggestion
        const pausedRegions = await pauseHandler.getPausedRegions();
```

test/protocol/PauseHandlerTest.js
```diff
@@ -125,16 +137,12 @@ describe("IBosonPauseHandler", function () {
...
-
         // Pause protocol again
-        await pauseHandler.connect(pauser).pause(regions);
+        await pauseHandler.connect(pauser).pause([PausableRegion.Twins]);
 
-        // Check that pauseScenario remains the sam
-        const newPauseScenario = await getStorageAt(pauseHandler.address, protocolStatusStorageSlotNumber);
-        expect(newPauseScenario, "Protocol not paused").to.equal(pauseScenario);
+        // Check that regions remains the same
+        const pausedRegions = await pauseHandler.getPauseStatus();
```
zajck, line 144:
```suggestion
        const pausedRegions = await pauseHandler.getPausedRegions();
```

test/protocol/PauseHandlerTest.js
```diff
@@ -93,7 +103,11 @@ describe("IBosonPauseHandler", function () {
 
         // Check that all regions are paused
         const pauseScenario = await getStorageAt(pauseHandler.address, protocolStatusStorageSlotNumber);
-        expect(BigNumber.from(pauseScenario).toNumber(), "Protocol not paused").to.equal(8191);
+        expect(BigNumber.from(pauseScenario).toNumber(), "Protocol not paused").to.equal(ALL_REGIONS_MASK);
+
+        // Check that all regions are paused
+        const pausedRegions = await pauseHandler.getPauseStatus();
```
zajck, line 109:
```suggestion
        const pausedRegions = await pauseHandler.getPausedRegions();
```

zajck (changes requested):

contracts/protocol/facets/PauseHandlerFacet.sol
anajuliabit: cool, applied!

test/protocol/PauseHandlerTest.js
anajuliabit: removed

contracts/protocol/facets/PauseHandlerFacet.sol
anajuliabit: slither action seems to be failing due to this change, and I haven't discovered the reason yet. I will search more on Monday.

anajuliabit: I have figured out that the problem is with the line https://github.com/bosonprotocol/boson-protocol-contracts/blob/c9ee232050034e9f12b0b17ef8c56466f0e5ba11/contracts/protocol/facets/PauseHandlerFacet.sol#L75
It appears that Slither is misinterpreting this line. [I have already submitted an issue on slither repo](https://github.com/crytic/slither/issues/2020).

While we wait for the bug to be fixed, we have two options to consider. First, we can temporarily disable the Slither action. Alternatively, we can hardcode the variable to 12, the current number of regions.

@zajck @mischat what do you guys prefer?

zajck: I think we can disable slither temporarily.
If there were any problems, it already raised warnings in other PRs.

zajck (approved):

zajck (approved):

mischat (approved):
LGTM  ... we will need this :)

PR #652, opened 2023-05-29
Fix unsuccessful twin transfer

zajck:
Fix #651

PR #653, opened 2023-05-29
Check that the twin address is a contract

### LINKED ISSUE #616: Fix code scanning alert - Low-level calls
Tracking issue for:
- [x] https://github.com/bosonprotocol/boson-protocol-contracts/security/code-scanning/1031

OZ Address library cannot be used to avoid these calls since we don't want to revert on unsuccessful calls.
But we can add check for code existence before making an external call.

zajck:
Fix #616 

This is branched from #652  which should be merged before this.

### DIFF
--- a/contracts/protocol/facets/ExchangeHandlerFacet.sol
+++ b/contracts/protocol/facets/ExchangeHandlerFacet.sol
@@ -722,43 +722,45 @@ contract ExchangeHandlerFacet is IBosonExchangeHandler, BuyerBase, DisputeBase {
                 }
 
                 // Calldata to transfer the twin
-                bytes memory data;
+                {
+                    bytes memory data;
 
-                if (tokenType == TokenType.FungibleToken) {
-                    // ERC-20 style transfer
-                    data = abi.encodeCall(IERC20.transferFrom, (seller.assistant, sender, twin.amount));
-                } else if (tokenType == TokenType.NonFungibleToken) {
-                    // Token transfer order is ascending to avoid overflow when twin supply is unlimited
-                    if (twin.supplyAvailable == type(uint256).max) {
-                        twin.tokenId++;
-                    } else {
-                        // Token transfer order is descending
-                        tokenId += twin.supplyAvailable;
+                    if (tokenType == TokenType.FungibleToken) {
+                        // ERC-20 style transfer
+                        data = abi.encodeCall(IERC20.transferFrom, (seller.assistant, sender, twin.amount));
+                    } else if (tokenType == TokenType.NonFungibleToken) {
+                        // Token transfer order is ascending to avoid overflow when twin supply is unlimited
+                        if (twin.supplyAvailable == type(uint256).max) {
+                            twin.tokenId++;
+                        } else {
+                            // Token transfer order is descending
+                            tokenId += twin.supplyAvailable;
+                        }
+
+                        // ERC-721 style transfer
+                        data = abi.encodeWithSignature(
+                            "safeTransferFrom(address,address,uint256,bytes)",
+                            seller.assistant,
+                            sender,
+                            tokenId,
+                            ""
+                        );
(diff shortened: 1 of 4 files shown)


contracts/protocol/facets/ExchangeHandlerFacet.sol
```diff
@@ -718,46 +721,46 @@ contract ExchangeHandlerFacet is IBosonExchangeHandler, BuyerBase, DisputeBase {
...
-                    // ERC-721 style transfer
-                    (success, result) = twin.tokenAddress.call(
-                        abi.encodeWithSignature(
+                // Calldata to transfer the twin
+                {
+                    bytes memory data;
+
+                    if (tokenType == TokenType.FungibleToken) {
+                        // ERC-20 style transfer
+                        data = abi.encodeCall(IERC20.transferFrom, (seller.assistant, sender, twin.amount));
```
mischat, line 730: I wonder why we decided to the change this one and not the subsequent similar calls on lines 742 and 751

mischat: ?

contracts/protocol/facets/ExchangeHandlerFacet.sol
```diff
@@ -718,46 +721,46 @@ contract ExchangeHandlerFacet is IBosonExchangeHandler, BuyerBase, DisputeBase {
...
+                    } else if (tokenType == TokenType.NonFungibleToken) {
+                        // Token transfer order is ascending to avoid overflow when twin supply is unlimited
+                        if (twin.supplyAvailable == type(uint256).max) {
+                            twin.tokenId++;
+                        } else {
+                            // Token transfer order is descending
+                            tokenId += twin.supplyAvailable;
```
mischat, line 737: This is just an aside, by i prefer the previous syntax here, it was more verbose, i like verbose as I have a simple internal parser ... but anyways, no need to change

contracts/protocol/facets/ExchangeHandlerFacet.sol
```diff
@@ -718,46 +721,46 @@ contract ExchangeHandlerFacet is IBosonExchangeHandler, BuyerBase, DisputeBase {
...
                             "safeTransferFrom(address,address,uint256,uint256,bytes)",
                             seller.assistant,
                             sender,
                             tokenId,
                             twin.amount,
                             ""
-                        )
-                    );
+                        );
+                    }
```
mischat, line 758: I guess also, that there is no way for us to have an unsupported TokenType at this point? If it is possible, would it be  sane to have an else clause and to throw a "unknown token type" kind of revert ? If so, we could always create an issue for the next release

mischat (approved):
I have left some comments, would be good to get responses, but i think in the worst case we just create an issue to fix later ... Looks good

contracts/protocol/facets/ExchangeHandlerFacet.sol
anajuliabit: there isn't. tokenType is defined when creating a twin, and if a seller passes a value that is not present in the TokenType enum transaction will be reverted.

anajuliabit (approved):
Great job 👍

contracts/protocol/facets/ExchangeHandlerFacet.sol
mischat: OK, thanks for the reply @anajuliabit

contracts/protocol/facets/ExchangeHandlerFacet.sol
mischat: Thanks

contracts/protocol/facets/ExchangeHandlerFacet.sol
mischat: Agreed, thanks

PR #675, opened 2023-06-08
Remove protocol gas limits

anajuliabit:
Fix https://github.com/bosonprotocol/boson-protocol-contracts/issues/668


anajuliabit:
> getAvailableFunds can run into a gas block limit if tokenList gets too long.

As far as I know, view functions don't run out of gas. Am I wrong?

zajck:
> > getAvailableFunds can run into a gas block limit if tokenList gets too long.
> 
> As far as I know, view functions don't run out of gas. Am I wrong?

Yes, they do. The gas is consumed the same as in other functions and is limited by the gas block limit. It's necessary because of the halting problem.
The only difference is that you don't pay for this gas if you make a direct external call to the contract.
But if you call a view function from another contract, then you also pay for all the gas it consumes.

This was remediated in #698 (there you can even find a PoC of how it runs out of gas in one of the older commits, before it was remmediated [489598b](https://github.com/bosonprotocol/boson-protocol-contracts/pull/698/commits/489598b5d5cf7d1b1ba3848aeb053e7c964e2cf2))

contracts/protocol/facets/ProtocolInitializationHandlerFacet.sol
```diff
@@ -122,7 +122,6 @@ contract ProtocolInitializationHandlerFacet is IBosonProtocolInitializationHandl
         uint256 _maxPremintedVouchers = abi.decode(_initializationData, (uint256));
         require(_maxPremintedVouchers != 0, VALUE_ZERO_NOT_ALLOWED);
         protocolLimits().maxPremintedVouchers = _maxPremintedVouchers;
-        emit MaxPremintedVouchersChanged(_maxPremintedVouchers, msgSender());
```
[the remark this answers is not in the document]
zajck: I think this still needs a proper resolution since this `initV2_2_0` is currently not doing what is expected from `initV2_2_0`.

Either we should remove "old" inits. 
Or probably simpler, just keep the event `MaxPremintedVouchersChanged` in `IBosonConfigEvents`, and keep all inits so we have "history" in one place. (I'm more for this option).

test/protocol/ProtocolInitializationHandlerTest.js
```diff
@@ -458,20 +467,6 @@ describe("ProtocolInitializationHandler", async function () {
       );
     });
 
-    it("Should emit MaxPremintedVouchersChanged event", async function () {
```
zajck, line 461: Just related to my other comment (in contracts/protocol/facets/ProtocolInitializationHandlerFacet.sol).
If we bring `MaxPremintedVouchersChanged` back, this test should also be put back.

zajck (changes requested):

zajck (approved):
LGTM :+1:

contracts/protocol/clients/voucher/BosonVoucher.sol
```diff
@@ -284,18 +277,14 @@ contract BosonVoucherBase is IBosonVoucher, BeaconClientBase, OwnableUpgradeable
...
         // Get the first token to burn
         uint256 start = (range.lastBurnedTokenId == 0) ? range.start : (range.lastBurnedTokenId + 1);
 
         // Get the last token to burn
-        uint256 end = range.start + range.minted;
+        uint256 end = start + _amount;
 
         // End should be greater than start
-        require(end > start, NOTHING_TO_BURN);
+        require(end > start && end <= range.start + range.minted, AMOUNT_EXCEEDS_RANGE_OR_NOTHING_TO_BURN);
```
mischat, line 287: I guess the intention was to use range.start here and not `range`

mischat: cc @anajuliabit and @zajck  ...

zajck: I think what @anajuliabit  did is correct.
`range.start + range.minted` is the highest `tokenId` that can be burned, so it's making sure that `end` is not above that.

mischat: yeah, thank both. I will resolve, i was just calling it out, as I wanted to make sure that it was intentional.

mischat (approved):
LGTM

zajck (approved):

PR #694, opened 2023-06-20
Limit gas when doing twin transfers

### LINKED ISSUE #687: Transfer twins should limit the gas it passes to twin contract when making a transfer
`transferTwins` sends all available gas to twin while doing twin transfers 
https://github.com/bosonprotocol/boson-protocol-contracts/blob/451dc3d0866ba6694821537803ba32939de34798/contracts/protocol/facets/ExchangeHandlerFacet.sol#L754-L761
https://github.com/bosonprotocol/boson-protocol-contracts/blob/451dc3d0866ba6694821537803ba32939de34798/contracts/protocol/facets/ExchangeHandlerFacet.sol#L771-L779
https://github.com/bosonprotocol/boson-protocol-contracts/blob/451dc3d0866ba6694821537803ba32939de34798/contracts/protocol/facets/ExchangeHandlerFacet.sol#L782-L791

The seller can create a twin that consumes all available gas on transfer. If they make two of them and bundle up them with an offer, buyers can commit to it, but `redeemVoucher` cannot succeed.
The buyer can only cancel the voucher, but then they lose `buyerCancelPenalty`

The seller can make malicious bundles with new offers or with the existing ones if nobody had committed to them yet.

## Recommendation

Limit how much gas can the external contract consume. This will still make twin transfer fail, but `redeemVoucher` will succeed and the exchange will go into dispute (as implemented in #652 )

Twin region of the protocol was paused to prevent twin creation until the fix is applied.

zajck:
Fix #687 

In the initial approach, I added a constant limit to every transfer. This could work in combination with `maxTwinPerBundle`. After I've done the estimates of what sensible values for gas limit and max number of twins could be, I decided we need a more flexible approach.

Instead of having a hard limit, the protocol passes almost all gas to the external contract, it just keeps the minimal gas needed to successfully execute the whole transaction.

More details are available in a separate doc:
https://docs.google.com/document/d/1lMik5Nl9ETUuiOGWpwyCW1lBQizzGSsjMEKyHrFud14/edit?usp=sharing


contracts/protocol/facets/ExchangeHandlerFacet.sol
```diff
@@ -757,14 +764,21 @@ contract ExchangeHandlerFacet is IBosonExchangeHandler, BuyerBase, DisputeBase {
...
-                    // Make call only if code at address exists
-                    if (twin.tokenAddress.isContract()) {
-                        (success, result) = twin.tokenAddress.call(data);
+                    // Make call only if there is enough gas and code at address exists.
+                    // If not, skip the call and mark the transfer as failed
+                    if (gasleft() > reservedGas && twin.tokenAddress.isContract()) {
+                        bytes memory result;
+                        (success, result) = twin.tokenAddress.call{ gas: gasleft() - reservedGas }(data);
+
+                        success = success && (result.length == 0 || abi.decode(result, (bool)));
```
mischat, line 773: Changing this around from the old line 767 was fun :)

mischat (approved):
This looks good to me, i left a message, but it isn't actionable.

contracts/protocol/facets/ExchangeHandlerFacet.sol
```diff
@@ -757,14 +764,21 @@ contract ExchangeHandlerFacet is IBosonExchangeHandler, BuyerBase, DisputeBase {
...
                     }
                 }
 
+                // Reduce minimum gas required for succesful execution
+                reservedGas -= SINGLE_TWIN_RESERVED_GAS;
+
                 // If token transfer failed
-                if (!success || (result.length > 0 && !abi.decode(result, (bool)))) {
+                if (!success) {
```
anajuliabit, line 781: why remove this? `|| (result.length > 0 && !abi.decode(result, (bool))`

anajuliabit (approved):
Looks really good 

just a note that the estimation needs to be evaluated again after https://github.com/bosonprotocol/boson-protocol-contracts/pull/656 get merged

contracts/protocol/facets/ExchangeHandlerFacet.sol
zajck: Wasn't removed, just turned around and moved to 773 (to solve stack too deep).

anajuliabit: nice

Issue #679, opened 2023-06-09
Polluting buyer's and agent's available funds


zajck:
#322 addresses the possibility to fill up the seller's available funds with unwanted tokens since anyone can deposit funds on the seller's behalf.

It's possible to do the same to buyers and agents as well. 

To fill up the buyer's available funds, a malicious actor can act as a seller and:
1. create an offer with an unwanted token and non-zero seller deposit;
2. commit on the buyer's behalf
3. revoke voucher, which releases unwanted token (seller deposit) to the buyer

To fill up the agent's available funds, a malicious actor can act as a seller and:
1. create an offer with an unwanted token and assign the agent to the offer;
2. commit to the offer;
3. redeem the voucher; 
4. finalize the exchange, which releases the unwanted token (percentage of the price) to the agent

- This fills up their available funds but does not send the (potentially malicious) token directly to the recipient. But the recipient can still withdraw it.
- This does not affect the withdrawal of other buyer's or agent's funds
- This pollutes `getAvailableFunds` with unwanted tokens
- Dispute resolvers are not affected, since they can receive DR fees only in the exchange tokens they specify.

Issue #322, opened 2022-09-12
FHF-01M: Inexistent Validation of Token Type


cliffhall:
## FHF-01M: Inexistent Validation of Token Type

| Type | Severity | Location |
| ---- | -------- | -------- |
| [Logical Fault](/appendix/finding-types#logical-fault) | <img class="o-severity o-minor" src="https://omniscia.io/report-assets/minor.png" /> | [FundsHandlerFacet.sol:L43-L67](https://github.com/bosonprotocol/boson-protocol-contracts/blob/25ea648255481178eb4d0821f973a6489b743a97/contracts/protocol/facets/FundsHandlerFacet.sol#L43-L67) |

### Description:

The `depositFunds` function permits an arbitrary caller to deposit funds available for a particular `_sellerId` with no form of validation on the caller or input arguments, thereby permitting elaborate phishing attacks to occur whereby a user invokes `depositFunds` with a malicious token to a particular user's account and when the target user invokes `withdrawFunds` with the automatic method on, the malicious token balance is automatically withdrawn to the user.

### Impact:

By depositing funds to a particular seller prior to f.e. their withdrawal transaction being processed, it is possible to "inject" a malicious token balance the user will withdraw seemingly from the Boson Protocol which will make the funds seem "trustworthy" and potential "bonuses" provided by Boson. This in turn will render phishing attacks easier to execute for unsuspecting users.

### Example:

```solidity
function depositFunds(
    uint256 _sellerId,
    address _tokenAddress,
    uint256 _amount
) external payable override fundsNotPaused nonReentrant {
    //Check Seller exists in sellers mapping
    (bool exists, , ) = fetchSeller(_sellerId);

    //Seller must exist
    require(exists, NO_SUCH_SELLER);

    if (msg.value != 0) {
        // receiving native currency
        require(_tokenAddress == address(0), NATIVE_WRONG_ADDRESS);
        require(_amount == msg.value, NATIVE_WRONG_AMOUNT);
    } else {
        // transfer tokens from the caller
        FundsLib.transferFundsToProtocol(_tokenAddress, _amount);
    }

    // increase available funds
    FundsLib.increaseAvailableFunds(_sellerId, _tokenAddress, _amount);

    emit FundsDeposited(_sellerId, msgSender(), _tokenAddress, _amount);
}
```

### Recommendation:

[...]

PR #698, opened 2023-06-22
More flexible getAvailableFunds

### LINKED ISSUE #546: Fix code scanning alert - Calls inside a loop
Tracking issue for:
- [ ] https://github.com/bosonprotocol/boson-protocol-contracts/security/code-scanning/66

### Description:

`getAvailableFunds` is a view function that returns the information about all funds that entity can withdraw. It accepts entity id as input argument and returns a list of structs containing “tokenAddress”, “tokenName” and “availableAmount”. For efficiency purposes "tokenName" is not stored  in the contract, but is fetched it on the fly every time getAvailableFunds is invoked. This is an external call to token contract. If method “name” does not exist on that contract (possible since it’s not part of ERC20 standard) or it reverts for some reason, it is handled and generic “Token name unspecified” is used in place of "tokenName" and the call still succeeds.

Although reverts are handled, an attacker can still try to break `getAvailableFunds` by making `name` consume all available resources.
Simplest way to achieve this is to make a custom token contract with “name” with an unbounded while loop inside. Although one malicious token does not immediatly break `getAvailableFunds`, combination of multiple does the trick.

`getAvailableFunds` is an external function and its purpose is to help users (or dApps) to easily retrieve all information needed to supply in `withdrawFunds` whenever they want to withdraw (part of) their funds. Malfunction of `getAvailableFunds` therefore presents the biggest issue to dApps.  
On the other hand, protocol is not severly affected. `getAvailableFunds` can be broken for individual users, but this does not affect others that don’t have malicious tokens on their token list.
Even if getAvailableFunds does not work, all funds are still safe and available to withdraw. It’s true that users cannot get the information about the maximum amount they can withdraw, but “withdrawFunds” still works normally.

### Recommendation:
The easiest way, which does not affect other part of the protocol and interfaces is to set manual limit when `name` is called, for example

```solidity
try IERC20Metadata(tokenAddress).name{gas: 100000}() returns (string memory name) 
```

zajck:
Fixes #546 

And also alleviates the consequences of #322 and #679


mischat:
and more merge conflicts here @zajck

mischat:
fwiw, I would have totally just removed the "name" field, and make that the client's problem, but i like the solution here nonetheless...

contracts/interfaces/handlers/IBosonFundsHandler.sol
```diff
@@ -72,8 +72,40 @@ interface IBosonFundsHandler is IBosonFundsEvents, IBosonFundsLibEvents {
...
+     *
+     * @param _entityId - id of entity for which availability of funds should be checked
+     * @return tokenList - list of token addresses
+     */
+    function getTokenList(uint256 _entityId) external view returns (address[] memory tokenList);
+
+    /**
+     * @notice Returns list of addresses for which the entity has funds available.
+     *
+     * @param _entityId - id of entity for which availability of funds should be checked
```
anajuliabit, line 87: add _limit and _offset parameters here

contracts/protocol/facets/FundsHandlerFacet.sol
```diff
@@ -149,24 +149,75 @@ contract FundsHandlerFacet is IBosonFundsHandler, ProtocolBase {
...
+     * @return tokenList - list of token addresses
+     */
+    function getTokenList(uint256 _entityId) external view override returns (address[] memory tokenList) {
+        return protocolLookups().tokenList[_entityId];
+    }
+
+    /**
+     * @notice Returns list of addresses for which the entity has funds available.
+     *
+     * @param _entityId - id of entity for which availability of funds should be checked
```
anajuliabit, line 166: add _limit and _offset parameters here

contracts/protocol/facets/FundsHandlerFacet.sol
```diff
@@ -189,6 +240,8 @@ contract FundsHandlerFacet is IBosonFundsHandler, ProtocolBase {
             availableFunds[i].tokenName = tokenName;
             availableFunds[i].availableAmount = entityFunds[tokenAddress];
         }
+
+        return availableFunds;
```
anajuliabit, line 244: no need for this return exist

anajuliabit (changes requested):
Looks good, just some minor suggestions

anajuliabit (approved):

mischat (approved):

PR #703, opened 2023-07-04
Clone vouchers using create2

### LINKED ISSUE #702: Use create2 instead of create in cloneBosonVoucher
Currently, the boson voucher clones are created by using `create` opcode.

https://github.com/bosonprotocol/boson-protocol-contracts/blob/ab3d4baf21db84b9fc7915abaf865f0ec4b607d5/contracts/protocol/bases/SellerBase.sol#L172

Therefore the new voucher address depends solely on the protocol contract address and its nonce. Although that's not a bug (except in a very unlikely scenario including a reorg and some specific racing conditions), using a `create2` would be an enhancement. It would allow sellers to have the boson vouchers with the same contract address on multiple chains once Boson protocol is present on multiple chains.

Recommendation:
Change `create` to `create2`, where the salt is derived from the seller assistant address and collection index. The exact derivation should take into account that the assistant's address can change. Probably some constant, derived form the seller address should be stored during the seller creation (for example `keccak256(sellerAddress)`).

zajck:
Closes #702 

Based on #592 which should be merged first.

Changes:
- use `create2` instead of `create` in `cloneBosonVoucher`
- `BeaconClientProxy` is deployed from the protocol.
- seller creator is stored and used when computing salt for additional collections


mischat (dismissed):
Looks good, thanks

contracts/protocol/bases/SellerBase.sol
```diff
@@ -142,36 +142,42 @@ contract SellerBase is ProtocolBase, IBosonAccountEvents {
...
+     * @param _creator - address of the seller creator
      * @param _assistant - address of the assistant
      * @param _voucherInitValues - the fully populated BosonTypes.VoucherInitValues struct
+     * @param _externalId - external collection id ("" for the default collection)
      * @return cloneAddress - the address of newly created clone
      */
     function cloneBosonVoucher(
         uint256 _sellerId,
         uint256 _collectionIndex,
+        address _creator,
```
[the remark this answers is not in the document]
mischat: thanks for responding to this offline

contracts/protocol/bases/SellerBase.sol
```diff
@@ -142,36 +142,42 @@ contract SellerBase is ProtocolBase, IBosonAccountEvents {
...
         address _assistant,
-        VoucherInitValues calldata _voucherInitValues
+        VoucherInitValues calldata _voucherInitValues,
+        string memory _externalId
```
mischat, line 165: keen to understand why you went for a string here :)

mischat (approved):
I am happy with this, i have a question which is merely informational for me

contracts/protocol/bases/SellerBase.sol
zajck: For the seller's convenience, I guess. 
I name the first collection "cats", the second "dogs", the third "frogs" etc.

It's not used as a key anywhere, so it's not that problematic.

But yes, we can make it bytes32 and be more efficient.

scripts/util/deploy-protocol-clients.js
```diff
@@ -36,7 +36,7 @@ async function deployProtocolClients(protocolClientArgs, maxPriorityFeePerGas, i
   // Deploy Protocol Client proxy contracts
   const protocolClientProxies = await deployProtocolClientProxies(protocolClientBeacons, maxPriorityFeePerGas);
 
-  // Cast the proxies to their implementation interfaces
+  // Cast the proxies to their implementation interfaces ?? ToDo: what is this even needed?
```
anajuliabit, line 39: That's a great question. I have checked all the places where we use this value and the correct seller BV address is calculated and attached before any calls are made. This function call (and consequently this cast-protocol-clients-proxies file) is unnecessary. Nice catch. (nothing saying to remove it here, just point it out for future revision)

contracts/protocol/bases/SellerBase.sol
mischat: thanks, no need to change, similar to the "getting the name for the ERC20s" i would have just used identifiers and given that responsibility to the clients ... 

In this instance, i would have said that the clients would have had to get the name from the metadata file, and the protocol woudl have only known about IDs .. but yeah, thanks

test/protocol/ProtocolInitializationHandlerTest.js
```diff
@@ -655,32 +658,100 @@ describe("ProtocolInitializationHandler", async function () {
...
+        // NEED TO ACTUALLY DEPLOY VOUCHER IMPLEMENTATIONS
+        const protocolClientArgs = [protocolDiamondAddress];
+        const [, beacons] = await deployProtocolClients(
+          protocolClientArgs,
+          maxPriorityFeePerGas,
+          [rando.address] // random address in place of forwarder
+        );
+        const [beacon] = beacons;
+
+        const facetsToDeploy = await getV2_2_0DeployConfig(); // To deploy 2.2.1, we can use 2.2.0 config
```
anajuliabit, line 681: I prefer to set this in the `00_config.js:getFacets` function as it is the designated location for storing all version configurations (which can then be reused for upgrade tests). but I am willing to approve this way as I can change it when I begin working on https://github.com/bosonprotocol/boson-protocol-contracts/pull/680.

test/protocol/ProtocolInitializationHandlerTest.js
```diff
@@ -655,32 +658,100 @@ describe("ProtocolInitializationHandler", async function () {
...
+                // set beacon proxy from config, not the deployed one
+                line = line.replace(
+                  "setBeaconProxyAddress(beaconProxy)",
+                  "setBeaconProxyAddress(_addresses.beaconProxy)"
+                );
+              }
+              return line;
+            },
+          }),
+        };
```
anajuliabit, line 709: nice workaround here 👍

test/protocol/ProtocolInitializationHandlerTest.js
```diff
@@ -743,6 +837,73 @@ describe("ProtocolInitializationHandler", async function () {
         ).to.be.revertedWith(RevertReasons.VALUE_ZERO_NOT_ALLOWED);
       });
 
+      it("sellerIds and sellerCreators length mismatch", async function () {
+        // set invalid minResolutionPeriod
```
anajuliabit, line 841:
```suggestion
```

test/protocol/ProtocolInitializationHandlerTest.js
```diff
@@ -743,6 +837,73 @@ describe("ProtocolInitializationHandler", async function () {
...
+            [facetCut],
+            deployedProtocolInitializationHandlerFacetAddress,
+            calldataProtocolInitialization,
+            await getFees(maxPriorityFeePerGas)
+          )
+        ).to.be.revertedWith(RevertReasons.ARRAY_LENGTH_MISMATCH);
+      });
+
+      it("invalid seller id ", async function () {
+        // set invalid minResolutionPeriod
```
anajuliabit, line 862:
```suggestion
```

test/protocol/ProtocolInitializationHandlerTest.js
```diff
@@ -743,6 +837,73 @@ describe("ProtocolInitializationHandler", async function () {
...
+            [facetCut],
+            deployedProtocolInitializationHandlerFacetAddress,
+            calldataProtocolInitialization,
+            await getFees(maxPriorityFeePerGas)
+          )
+        ).to.be.revertedWith(RevertReasons.NO_SUCH_SELLER);
+      });
+
+      it("invalid seller creator address ", async function () {
+        // set invalid minResolutionPeriod
```
anajuliabit, line 885:
```suggestion
```

test/util/utils.js
```diff
@@ -251,6 +253,40 @@ function calculateContractAddress(senderAddress, senderNonce) {
...
+  const contract_address = "0x" + contract_address_long.substring(26); //Trim the first 24 characters.
+
+  return getAddress(contract_address);
+}
+
+function getCloneByteCodeHash(beaconProxyAddress) {
+  return keccak256(
+    `0x3d602d80600a3d3981f3363d3d373d3d3d363d73${beaconProxyAddress.slice(2)}5af43d82803e903d91602b57fd5bf3`
+  );
+}
```
anajuliabit, line 271: where did you get this value? would be nice to have some comments here :)

anajuliabit (changes requested):
Looks good. I added some suggestions, but nothing that has an impact on contracts.

test/protocol/ProtocolInitializationHandlerTest.js
mischat: do you want to add a note so we don't lose this one ? @anajuliabit

anajuliabit (approved):

test/protocol/ProtocolInitializationHandlerTest.js
anajuliabit: done

anajuliabit (approved):

test/util/utils.js
zajck: I got it from our contracts, but otherwise this is [EIP 1167](https://eips.ethereum.org/EIPS/eip-1167) style minimal clone.

We use the code from [this example](https://github.com/optionality/clone-factory/blob/master/contracts/CloneFactory.sol).

OZ uses the same: https://blog.openzeppelin.com/deep-dive-into-the-minimal-proxy-contract

PR #718, opened 2023-07-27
Replace approve for forceApprove on BosonVoucher

### LINKED ISSUE #701: ERC20 approvals in bosonVoucher
There are two problems, related to approving ERC20 transfers for tokens owned by the boson voucher contract.

1. Incompatibility with non-ERC20-compliant tokens
   Some tokens on mainnet are not  100% compliant with ERC20 standard (USDT for example). Since their method `approve` returns nothing, it makes `withdrawToProtocol` unusable for these tokens:  
https://github.com/bosonprotocol/boson-protocol-contracts/blob/ab3d4baf21db84b9fc7915abaf865f0ec4b607d5/contracts/protocol/clients/voucher/BosonVoucher.sol#L785-L787
   **Recommendation**: The method `safeIncreaseAllowance` from OZ's safeERC20 library handles this, so it's recommended to be implemented here.
2. Assistant can transfer contract-owned tokens, by using non-standard `increaseAllowance`
   In `callExternalContract` we prevent the invocation of certain selectors, to prevent transferring the funds from the voucher contract.   
https://github.com/bosonprotocol/boson-protocol-contracts/blob/ab3d4baf21db84b9fc7915abaf865f0ec4b607d5/contracts/protocol/clients/voucher/BosonVoucher.sol#L545-L552
Currently, this list includes standard ERC20 methods and DAI's aliases for these methods. However, a lot of tokens nowadays implement non-standard `increaseAllowance` which can be used to achieve the same effect as `approve`, i.e. assistant can approve any address that can then transfer tokens on the contract's behalf.
   **Recommendation**: add `increaseAllowance` to the list of prohibited methods.

anajuliabit:
Close #701 (only first item)


contracts/protocol/clients/voucher/BosonVoucher.sol
```diff
@@ -774,7 +776,7 @@ contract BosonVoucherBase is IBosonVoucher, BeaconClientBase, OwnableUpgradeable
                 IBosonFundsHandler(protocolDiamond).depositFunds{ value: balance }(sellerId, token, balance);
             } else {
                 uint256 balance = IERC20(token).balanceOf(address(this));
-                IERC20(token).approve(protocolDiamond, balance);
+                IERC20(token).safeIncreaseAllowance(protocolDiamond, balance);
```
levalleux-ludo, line 779: In order to keep same behavior
```suggestion
                currentAllowance = ...;
                if (currentAllowance < balance) {
                    IERC20(token).safeIncreaseAllowance(protocolDiamond, balance - currentAllowance);
               }
```

levalleux-ludo (approved):
Left 1 remark (approved anyway)

contracts/protocol/clients/voucher/BosonVoucher.sol
anajuliabit: done

contracts/protocol/clients/voucher/BosonVoucher.sol
```diff
@@ -774,7 +776,12 @@ contract BosonVoucherBase is IBosonVoucher, BeaconClientBase, OwnableUpgradeable
                 IBosonFundsHandler(protocolDiamond).depositFunds{ value: balance }(sellerId, token, balance);
             } else {
                 uint256 balance = IERC20(token).balanceOf(address(this));
-                IERC20(token).approve(protocolDiamond, balance);
+                // get current allowance
+                uint256 allowance = IERC20(token).allowance(address(this), protocolDiamond);
+                if (allowance < balance) {
+                    // increase allowance if needed
+                    IERC20(token).safeIncreaseAllowance(protocolDiamond, balance - allowance);
```
zajck, line 783:
```suggestion
                    IERC20(token).forceApprove(protocolDiamond, balance);
```

Just use this directly to reduce the number of external calls.

contracts/protocol/clients/voucher/BosonVoucher.sol
```diff
@@ -774,7 +776,12 @@ contract BosonVoucherBase is IBosonVoucher, BeaconClientBase, OwnableUpgradeable
                 IBosonFundsHandler(protocolDiamond).depositFunds{ value: balance }(sellerId, token, balance);
```
zajck, line 776:
```suggestion
                IBosonFundsHandler(protocolDiamond).depositFunds{ value: balance }(sellerId, token, balance);
```
Not directly connected to this PR, but I guess we could add this to prevent unnecessary calls.

contracts/protocol/clients/voucher/BosonVoucher.sol
```diff
@@ -774,7 +776,12 @@ contract BosonVoucherBase is IBosonVoucher, BeaconClientBase, OwnableUpgradeable
...
             } else {
                 uint256 balance = IERC20(token).balanceOf(address(this));
-                IERC20(token).approve(protocolDiamond, balance);
+                // get current allowance
```
zajck, line 779: Similar to above, we can check if the balance  > 0 and do the other operations only then.

test/protocol/FundsHandlerTest.js
```diff
@@ -367,12 +367,15 @@ describe("IBosonFundsHandler", function () {
...
           await foreign20ReturnFalse.connect(assistant).approve(protocolDiamondAddress, depositAmount);
+          const safeERC20 = await ethers.getContractAt("SafeERC20", protocolDiamondAddress);
 
           await expect(
             fundsHandler
               .connect(assistant)
               .depositFunds(seller.id, await foreign20ReturnFalse.getAddress(), depositAmount)
-          ).to.revertedWith(RevertReasons.SAFE_ERC20_NOT_SUCCEEDED);
+          )
+            .to.revertedWithCustomError(safeERC20, "SafeERC20FailedOperation")
```
zajck, line 377: I suppose we should also put custom errors in the revert-reasons.js for consistency.

zajck (changes requested):
Another remark regarding `ext_libs`.

I think we have them since Cliff wanted protocol dependencies to be "static", i.e. what protocol imports is part of the codebase and not imported from OZ package directly.

Personally, I don't think that's necessary, especially since we'll have to update our "ext_libs" from time to time anyway (in cases like this PR). So I think it's fine if we get rid of `contracts/ext_libs/SafeERC20.sol` file and just import OZ version wherever is needed.
Thoughts? @anajuliabit @mischat

anajuliabit:
@zajck applied everthing

zajck (approved):
LGTM :+1:

albertfolch-redeemeum (approved):
LGTM

PR #759, opened 2023-08-07
Raise dispute only once even if multiple twin transfers fail

### LINKED ISSUE #758: If multiple twin transfer fail, dispute is raised multiple times
Inside `transferTwins`, `raiseDisputeInternal` is called inside a loop (lines 811-923), directly after an unsuccessful twin transfer

https://github.com/bosonprotocol/boson-protocol-contracts/blob/c1952c58acbf093b6424e8119296f02d0bf9eaeb/contracts/protocol/facets/ExchangeHandlerFacet.sol#L880

If multiple twin transfers fail, `raiseDisputeInternal` is also called multiple times. Although not directly exploitable, it uses the gas and pollutes the logs with multiple copies of the same event.

It needs to be moved outside the loop and the dispute should be raised only once.

zajck:
fix #758 

Fix required some additional refactoring to avoid stack too deep error.


albertfolch-redeemeum (dismissed):
LGTM

albertfolch-redeemeum (approved):

mischat (approved):
This seems sane. thanks

PR #771, opened 2023-08-11
Remediate malformed return during the twin transfer

### LINKED ISSUE #769: twin transfer issue - malformed return data
Despite the protective measures currently implemented, the seller can still make `redeemVoucher` to revert and effectively force the buyer to either cancel the voucher or let it expire (in either case, the buyer loses the cancellation penalty).

There are possible attacks:
- returnbomb attack (#768)
- malformed return data (this issue)

## Maformed return data

If the twin transfer succeeds, the protocol will try to decode the result (if its length is >0)
https://github.com/bosonprotocol/boson-protocol-contracts/blob/9070c0484634cec7bec793808a43ca20616d889a/contracts/protocol/facets/ExchangeHandlerFacet.sol#L871

If the result is malformed (too long, too short or contains a value that cannot be bool (e.g. 3)), the transaction will revert without a revert reason.

Example:
https://github.com/bosonprotocol/boson-protocol-contracts/blob/a4949ec7969df0d3779136bd9eaa7ddfef8f116b/contracts/mock/Foreign20.sol#L266-L278

### Recommendation

Manually decode the result, i.e. check that the return data is 32 bytes long and accept 1 as a success and anything else as a transfer failure.

zajck:
Fix #769 

Try-catch can be used only on external calls/contract creation, therefore it needed to be handled differently.

The most concise way (and without the need for assembly) was to say that return is valid iff its length is 0 or length is 32 and decoded into uint256 its value is exactly 1 (corresponds to `true` if considered a boolean).


mischat:
@zajck could you please fix the merge conflicts in this PR ?

albertfolch-redeemeum (approved):
LGTM

PR #772, opened 2023-08-11
Remediate returnbomb during a twin transfer

### LINKED ISSUE #768: twin transfer issue - returnbomb attack
Despite the protective measures currently implemented, the seller can still make `redeemVoucher` to revert and effectively force the buyer to either cancel the voucher or let it expire (in either case, the buyer loses the cancellation penalty).

There are possible attacks:
- returnbomb attack (this issue)
- malformed return data (#769) 

## Returnbomb attack

The twin transfer reverts with insanely long revert data. The return value gets copied into the memory here
https://github.com/bosonprotocol/boson-protocol-contracts/blob/9070c0484634cec7bec793808a43ca20616d889a/contracts/protocol/facets/ExchangeHandlerFacet.sol#L869

If the result is of the right size it can consume enough gas to make the transaction fail.

Example:
https://github.com/bosonprotocol/boson-protocol-contracts/blob/a4949ec7969df0d3779136bd9eaa7ddfef8f116b/contracts/mock/Foreign20.sol#L251-L263

### Recommendation

Use a solution like this https://github.com/nomad-xyz/ExcessivelySafeCall which limits the amount of return data copied to the memory.

zajck:
Fixes #768

Currently, the only approach was to use the assembly, which allows ignoring long return data.


albertfolch-redeemeum (dismissed):
LGTM

mischat:
@zajck ^^

albertfolch-redeemeum (approved):

mischat (approved):
what a find this one is, it is very interesting, and neat to understand how this happens. 

thanks for this, the assembly looks good and circumvents the return bombing!

PR #781, opened 2023-08-24
Dry run option for suite deployment

zajck:
So far, only the "migrate" script had the option to simulate it without an actual deployment.

This PR adds the same option for the initial deployment.


levalleux-ludo (approved):
LGTM

mischat (approved):
looks good

PR #788, opened 2023-08-30
register commitToConditionalOffer method for meta-transactions

### LINKED ISSUE #787: Add support for EIP712 for `commitToConditionalOffer`
All methods used by buyers have a custom EIP712 structure which allows greater legibility for the buyers when they use metatransactions.

Recently introduced `commitToConditionalOffer` which will also be used by buyers does not have a custom EIP712 structure yet, meaning it can be invoked only as a generic metatransaction which makes it harder to understand what user is signing.

### Recommendation

Make `commitToConditionalOffer` a special metaTx function and add support for its specific EIP712 structure.

levalleux-ludo:
fixes #787

### DIFF
--- a/contracts/domain/BosonTypes.sol
+++ b/contracts/domain/BosonTypes.sol
@@ -64,9 +64,8 @@ contract BosonTypes {
         CommitToOffer,
         Exchange,
         Funds,
-        RaiseDispute,
-        ResolveDispute,
-        CommitToConditionalOffer
+        CommitToConditionalOffer,
+        ResolveDispute
     }
 
     enum AuthTokenType {
--- a/contracts/domain/BosonTypes.sol
+++ b/contracts/domain/BosonTypes.sol
@@ -65,7 +65,8 @@ contract BosonTypes {
         Exchange,
         Funds,
         RaiseDispute,
-        ResolveDispute
+        ResolveDispute,
+        CommitToConditionalOffer
     }
 
     enum AuthTokenType {
--- a/contracts/protocol/facets/MetaTransactionsHandlerFacet.sol
+++ b/contracts/protocol/facets/MetaTransactionsHandlerFacet.sol
@@ -31,7 +31,7 @@ contract MetaTransactionsHandlerFacet is IBosonMetaTransactionsHandler, Protocol
 
         // Set input type for the function name
         pmti.inputType[COMMIT_TO_OFFER] = MetaTxInputType.CommitToOffer;
-        pmti.inputType[COMMIT_TO_CONDITIONAL_OFFER] = MetaTxInputType.CommitToOffer;
+        pmti.inputType[COMMIT_TO_CONDITIONAL_OFFER] = MetaTxInputType.CommitToConditionalOffer;
         pmti.inputType[WITHDRAW_FUNDS] = MetaTxInputType.Funds;
         pmti.inputType[RESOLVE_DISPUTE] = MetaTxInputType.ResolveDispute;
         pmti.inputType[CANCEL_VOUCHER] = MetaTxInputType.Exchange;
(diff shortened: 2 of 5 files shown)


contracts/protocol/facets/MetaTransactionsHandlerFacet.sol
```diff
@@ -31,6 +31,7 @@ contract MetaTransactionsHandlerFacet is IBosonMetaTransactionsHandler, Protocol
 
         // Set input type for the function name
         pmti.inputType[COMMIT_TO_OFFER] = MetaTxInputType.CommitToOffer;
+        pmti.inputType[COMMIT_TO_CONDITIONAL_OFFER] = MetaTxInputType.CommitToOffer;
```
zajck, line 34: A new `MetaTxInputType` is needed for `COMMIT_TO_CONDITIONAL_OFFER`.

`MetaTxInputType.CommitToOffer` does not include `tokenId` parameter, so the buyer would be signing only `buyer` and `offerId`. Therefore the relayer can choose an arbitrary `tokenId` and the metaTx will succeed (as long as the buyer actually owns it). We need `MetaTxInputType.ConditionalCommitToOffer` which includes the `tokenId`.

To fully support it, we also need a hashing method `hashConditionalOfferDetails` which will correctly hash the inputs.

zajck (changes requested):
The change I requested will also affect the unit tests. I'll review them after the contracts are updated.

contracts/protocol/facets/MetaTransactionsHandlerFacet.sol
levalleux-ludo: ok

contracts/domain/BosonTypes.sol
```diff
@@ -65,7 +65,8 @@ contract BosonTypes {
         Exchange,
         Funds,
         RaiseDispute,
-        ResolveDispute
+        ResolveDispute,
+        CommitToConditionalOffer
```
zajck, line 69: You could actually replace `RaiseDispute` value with `CommitToConditionalOffer`.

`RaiseDispute` is not an actual type anymore, but we just kept it in for consistency of other values.
But now we have a meaningful replacement so we can get rid of it.

zajck (changes requested):
Looks good, just one small remark.

contracts/domain/BosonTypes.sol
levalleux-ludo: done

zajck (approved):
Looks good :+1:

contracts/domain/BosonTypes.sol
```diff
@@ -64,7 +64,7 @@ contract BosonTypes {
         CommitToOffer,
         Exchange,
         Funds,
-        RaiseDispute,
```
albertfolch-redeemeum, line 67: why no raise dispute?

test/util/upgrade.js
```diff
@@ -1098,7 +1099,7 @@ async function getMetaTxPrivateContractState(protocolDiamondAddress) {
     CommitToOffer: 1,
     Exchange: 2,
     Funds: 3,
-    RaiseDispute: 4,
```
albertfolch-redeemeum, line 1101: same here

contracts/domain/BosonTypes.sol
levalleux-ludo: RaiseDispute is obsolete, not used anymore. It was not removed because it would have create a "hole" in the enum values.
Now we have a new value in the enum so, instead of adding a new label/value, we replace this old one with the new one

albertfolch-redeemeum (approved):

contracts/domain/BosonTypes.sol
levalleux-ludo: see Klemen comment here: https://github.com/bosonprotocol/boson-protocol-contracts/pull/788#discussion_r1311279105

PR #790, opened 2023-09-01
Conditional commit eligibility

### LINKED ISSUE #786: View function to check eligibility for conditional offer
Users can now check if they are eligible to commit to a conditional offer only by estimating the gas of `commitToConditionalOffer` method.

Add a view method to get the information about eligibility and the remaining number of commits.

zajck:
Close #786 

- Added an event `ConditionalCommitAuthorized` which is emitted every time a buyer commits to conditional offer
- Added `isEligibleToCommit` which returns the eligibility status and the commits status


levalleux-ludo (approved):

albertfolch-redeemeum (approved):
LGTM

PR #794, opened 2023-09-06
Fix tokenRange removal

### LINKED ISSUE #792: When `tokenRange` is removed, `rangeIdByTwin` is not updated correctly
There are two cases when an element is removed from `_lookups.twinRangesBySeller` array:
- in `updateNFTRanges`
- in `removeTwin` if the twin token type is `NonFungibleToken`

If a seller has multiple twins with the same token, removal of token range is done with typical pattern, where we swap the last element and element that needs to be removed:

https://github.com/bosonprotocol/boson-protocol-contracts/blob/7c271ee1d48cedb49f4a0067144a5afdeb5484df/contracts/protocol/facets/ExchangeHandlerFacet.sol#L1268-L1271

https://github.com/bosonprotocol/boson-protocol-contracts/blob/7c271ee1d48cedb49f4a0067144a5afdeb5484df/contracts/protocol/facets/TwinHandlerFacet.sol#L89-L91

This is fine, but at the same time `lookups.rangeIdByTwin` should be updated, too, since it's used to efficiently retrieve range id based on token id. Because we don't do it, the second removal of the twin with the same address can:
- remove the wrong twin or
- access the out of bond index which reverts the tx with `reverted with panic code 0x32 (Array accessed at an out-of-bounds or negative index)`. This can be even exploited to DoS `redeemVoucher` method.

### Recommendation

In both cases also update `lookups.rangeIdByTwin` when the swap is done.
To do it, one needs the twin id associated with the last token range. 
Currently, there is no efficient way to get it, therefore I propose to add a field `twinId` to the struct `TokenRange`

zajck:
Fix #792

### DIFF
--- a/contracts/domain/BosonTypes.sol
+++ b/contracts/domain/BosonTypes.sol
@@ -238,6 +238,7 @@ contract BosonTypes {
     struct TokenRange {
         uint256 start;
         uint256 end;
+        uint256 twinId;
     }
 
     struct Twin {
--- a/contracts/protocol/bases/TwinBase.sol
+++ b/contracts/protocol/bases/TwinBase.sol
@@ -92,6 +92,7 @@ contract TwinBase is ProtocolBase, IBosonTwinEvents {
             TokenRange storage tokenRange = twinRanges.push();
             tokenRange.start = tokenId;
             tokenRange.end = lastTokenId;
+            tokenRange.twinId = twinId;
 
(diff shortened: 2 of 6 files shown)


albertfolch-redeemeum (approved):
LGTM

mischat (approved):
This makes sense.

PR #800, opened 2023-09-07
Limit the number of twin transfers

### LINKED ISSUE #793: A large number of twins can make redemption impossibe
The number of twins in a bundle determines how many times the following loop will be executed:
https://github.com/bosonprotocol/boson-protocol-contracts/blob/7c271ee1d48cedb49f4a0067144a5afdeb5484df/contracts/protocol/facets/ExchangeHandlerFacet.sol#L870
https://github.com/bosonprotocol/boson-protocol-contracts/blob/7c271ee1d48cedb49f4a0067144a5afdeb5484df/contracts/protocol/facets/ExchangeHandlerFacet.sol#L880

A sufficiently large number of twins can make transactions run out of gas despite all the measures we currently have. Tests show that this can be done with approximately 550 twins, possibly less in a more advanced exploit. Although the number is high, it's still totally feasible.
As a consequence, the seller can DoS the `redeemVoucher` method, which forces the buyer to lose the cancellation penalty.

### Recommendations

1. Bring back the removed `maxTwinsPerBundle`. What would be an appropriate limit?
2. Compare `reservedGas` to `block.gaslimit`. If `reservedGas` is higher, don't even try to transfer anything and just raise a dispute. Still, if we want to issue `TwinTransferFailed` for all twins, there probably exists a number of twins where it starts to fail again.
3. Totally change the whole approach to handle twin transfer. Currently, users can raise a dispute only after they redeem the voucher. If a twin transfer fails, they cannot redeem it and consequently cannot raise a dispute. This could be changed so that for offers with twins users could raise a dispute even if they did not redeem the voucher yet. Needs further assessments of all impacts on the protocol.

zajck:
Fix #793 

This change skips the twin transfers completely if the reserved gas is higher than block gas limit.
It affects the bundles with more than 187 twins.
In normal usage, we expect only a few twins per bundle, so this change does not pose a big limitation.


test/protocol/ExchangeHandlerTest.js
```diff
@@ -5424,6 +5586,79 @@ describe("IBosonExchangeHandler", function () {
             // It should match ExchangeState.Disputed
             assert.equal(response, ExchangeState.Disputed, "Exchange state is incorrect");
           });
+
+          it("Too many twins", async function () {
```
albertfolch-redeemeum, line 5590:
```suggestion
          it("Too many twins - erc20 & erc721 twins", async function () {
```

albertfolch-redeemeum (approved):

test/protocol/ExchangeHandlerTest.js
zajck: I don't think this is necessary - this test is inside the context `Offer bundled with mixed twins` where all tests include a combination of different twins.
And it also uses erc115 twins.

mischat (approved):
For what it is worth this looks good, github ui made a real mess of that diff IMO... As it was merely moving soem code around and adding in a new if statement.

PR #821, opened 2023-11-07
Custom errors

### LINKED ISSUE #818: Replace revert reasons with custom errors
We use `require` statement more than 350 times in our contracts.
String revert reasons are quite expensive and add inflate the contract size, which can make them impossible to deploy without refactoring.

By changing them with custom error messages, we can optimize the gas costs for both deployment and consequent use.

zajck:
Close #818 

Whenever it makes sense, old string revert reasons were replaced with custom errors, which are more gas-efficient.

Some of the errors are now less verbose than before. The mapping between old errors and new ones can be found by comparing values in `scripts/config/revert-reasons-old.js` and `scripts/config/revert-reasons.js

There were some inconsistencies in the error naming style, especially with the ones containing the word "invalid". The majority of them had "invalid" as a prefix, while some had it as a suffix. This was now unified, so all have "invalid" at the beginning.

Size change:
image


levalleux-ludo (commented):
Adding NatSpec comments in contracts/domain/BosonErrors.sol could make sense

levalleux-ludo (dismissed):
LGTM

levalleux-ludo (approved):

albertfolch-redeemeum (approved):

PR #824, opened 2023-11-21
Add get getSellersCollectionsPaginated method

### LINKED ISSUE #709: Add getSellerCollection method
If a seller has numerous collections, `getSellersCollections` may run out of gas
We should create a getSellerCollection which has an additional parameter `collectionId`

https://github.com/bosonprotocol/boson-protocol-contracts/pull/592/files#r1254669259

zajck:
Closes #709 

Implementation is slightly different than proposed in the issue. It matches the pattern in the existing `getTokenListPaginated`.
This allows to retrieval of multiple collections with a single call, but with more granular control.

Changes:
- add `getSellersCollectionsPaginated` to `SellerHandlerFacet`
- add corresponding unit tests

### DIFF
--- a/contracts/interfaces/handlers/IBosonAccountHandler.sol
+++ b/contracts/interfaces/handlers/IBosonAccountHandler.sol
@@ -9,7 +9,7 @@ import { IBosonAccountEvents } from "../events/IBosonAccountEvents.sol";
  *
  * @notice Handles creation, update, retrieval of accounts within the protocol.
  *
- * The ERC-165 identifier for this interface is: 0xbc1d7461
+ * The ERC-165 identifier for this interface is: 0x8946080b
  */
 interface IBosonAccountHandler is IBosonAccountEvents {
     /**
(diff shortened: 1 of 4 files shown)


contracts/interfaces/handlers/IBosonAccountHandler.sol
```diff
@@ -450,7 +450,8 @@ interface IBosonAccountHandler is IBosonAccountEvents, BosonErrors {
     ) external view returns (bool exists, BosonTypes.Seller memory seller, BosonTypes.AuthToken memory authToken);
 
     /**
-     * @notice Gets the details about a seller's collections.
+     * @notice Gets the details about all seller's collections.
+     * In case seller has too many collections and this runs out of gas, pleas use getSellersCollectionsPaginated.
```
albertfolch-redeemeum, line 454:
```suggestion
     * In case seller has too many collections and this runs out of gas, please use getSellersCollectionsPaginated.
```

albertfolch-redeemeum: same typo in other places

albertfolch-redeemeum (approved):

levalleux-ludo (commented):
It would make sense to provide a method `getCollectionCount(sellerId) returns uint256`, IMHO

levalleux-ludo (approved):

PR #827, opened 2023-11-30
Update solidity version and remove event definition duplicates

### LINKED ISSUE #773: Remove duplicated events from libraries
Once #763 is closed, we can take advantage of the new feature introduced in 0.8.21 - "Qualified access to foreign events".

Some events are currently duplicated in libraries and interfaces, but now we can get rid of those in libraries and use the interface definitions instead.

More info: https://soliditylang.org/blog/2023/07/19/solidity-0.8.21-release-announcement/

List of duplicated events:
```
FundsReleased(uint256,uint256,address,uint256,address)
ProtocolFeeCollected(uint256,address,uint256,address)
FundsEncumbered(uint256,address,uint256,address)
FundsReleased(uint256,uint256,address,uint256,address)
ProtocolFeeCollected(uint256,address,uint256,address)
FundsWithdrawn(uint256,address,address,uint256,address)
```

Additionally, if we deprecate some events, but still need to refer to them for some legacy reasons, we could have a special interface containing only the deprecated events. Then we would invoke them using the new syntax. For example, this applies to `MaxPremintedVouchersChanged` which is not used anymore, but we still have it in `ProtocolInitializationHandlerFacet`

### LINKED ISSUE #826: Update solidity version
The latest [solidity release](https://github.com/ethereum/solidity/releases) is 0.8.23.

The highest version [supported by Hardhat](https://hardhat.org/hardhat-runner/docs/reference/solidity-support) is 0.8.22

It's necessary to do it to close #773 
It makes sense to do it for 2.4.0 since the change of error messages requires an upgrade of all contracts anyway.

zajck:
Close #773 
Close #826


scripts/util/detect-changed-contracts.js
```diff
@@ -34,7 +34,7 @@ async function detectChangedContract(referenceCommit, targetCommit = "HEAD") {
       transform: (line) => {
```
levalleux-ludo, line 34: I guess the comment 3 lines above (^^) does not match the change (still mention version 0.8.9 instead of 0.8.21

levalleux-ludo (approved):

PR #888, opened 2024-01-09
Minor seller handler optimization

zajck:
More efficient array length management, which became possible to implement after the changes in #883.

### DIFF
--- a/contracts/protocol/bases/PriceDiscoveryBase.sol
+++ b/contracts/protocol/bases/PriceDiscoveryBase.sol
@@ -31,6 +31,7 @@ contract PriceDiscoveryBase is ProtocolBase {
      */
     //solhint-disable-next-line
     constructor(address _wNative) {
+        if (_wNative == address(0)) revert InvalidAddress();
         wNative = IWrappedNative(_wNative);
     }
 
--- a/test/protocol/PriceDiscoveryHandlerFacet.js
+++ b/test/protocol/PriceDiscoveryHandlerFacet.js
@@ -152,6 +152,17 @@ describe("IPriceDiscoveryHandlerFacet", function () {
     });
   });
 
+  context("📋 Constructor", async function () {
+    it("Deployment fails if wrapped native address is 0", async function () {
+      const priceDiscoveryFactory = await getContractFactory("PriceDiscoveryHandlerFacet");
+
+      await expect(priceDiscoveryFactory.deploy(ZeroAddress)).to.revertedWithCustomError(
+        bosonErrors,
+        RevertReasons.INVALID_ADDRESS
+      );
+    });
+  });
+
   // All supported Price discovery methods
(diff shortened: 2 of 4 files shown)


albertfolch-redeemeum (approved):

contracts/protocol/facets/SellerHandlerFacet.sol
```diff
@@ -504,7 +505,7 @@ contract SellerHandlerFacet is SellerBase {
                 revert InvalidRoyaltyPercentage();
 
             royaltyRecipients.push(_royaltyRecipients[i]);
-            royaltyRecipientIndexByRecipient[_royaltyRecipients[i].wallet] = royaltyRecipients.length; // can be optimized to use counter instead of array length
+            royaltyRecipientIndexByRecipient[_royaltyRecipients[i].wallet] = royaltyRecipientsStorageLength + i;
```
gunnigylfa, line 508: Just curious. Before All elements of this mapping were getting the same value e.g. `royaltyRecipients.length` but now they are getting an ever increasing dynamic value starting from `royaltyRecipients.length + 1`
So two questions really: 
1. Are there any implications of the recipient index not being static anymore?
2. If not, are there any implications of starting the count from length + 1 instead of starting from the length

zajck: Even before, the elements were getting different values in this mapping.
`royaltyRecipients` is a storage reference (see line 490), so when we push an element to it in line 580, its length increases, which is reflected in `royaltyRecipients.length`.

So, the values of `royaltyRecipientIndexByRecipient` do not change after this change. The difference is just that the new code is a tiny bit more efficient. In old code `royaltyRecipients.length` requires SLOAD operation (i.e. reading the value from storage), while in the new implementation `royaltyRecipientsStorageLength + i` is using MLOAD (reading from the memory), which is less costly. In general, we want to reduce the reading from and writing to storage as much as possible.

The purpose of `royaltyRecipientIndexByRecipient` is to have an efficient lookup in other places in the code. There are two reasons we start with length + 1:
1. For the purpose of `royaltyRecipientIndexByRecipient`, `royaltyRecipients` is 1-based, not 0-based. The reason for this is that solidity does not differentiate between `0` and `null` values. If we used 0-based indexing, we could not differentiate between the recipient that is actually at index 0 and some address that is not even a recipient.
2. In line 491, `royaltyRecipients.length` represents the 1-based index of the last element of that array. When we in line 507 add another element, the length increases by 1, which should be reflected in the value that we store. Technically, we could omit `+ 1` in line 491, but then we'd need to add it in 508. But since `1` is constant, it makes sense to add it to the value outside the loop to not repeat the operation.

gunnigylfa: Got it, thank you for clarifying!

PR #894, opened 2024-01-19
[PDB-01M] Arbitrary External Contract Calls

### LINKED ISSUE #861: [PDB-01M] Arbitrary External Contract Calls
## PDB-01M: Arbitrary External Contract Calls

| [Logical Fault](/appendix/finding-types#logical-fault) |  | [PriceDiscoveryBase.sol:L132](https://github.com/bosonprotocol/boson-protocol-contracts/blob/b29767e7be7b14ff87e69fb2709b4324956b93df/contracts/protocol/bases/PriceDiscoveryBase.sol#L132), [L210](https://github.com/bosonprotocol/boson-protocol-contracts/blob/b29767e7be7b14ff87e69fb2709b4324956b93df/contracts/protocol/bases/PriceDiscoveryBase.sol#L210), [L275](https://github.com/bosonprotocol/boson-protocol-contracts/blob/b29767e7be7b14ff87e69fb2709b4324956b93df/contracts/protocol/bases/PriceDiscoveryBase.sol#L275) |

### Description:

The `PriceDiscoveryBase` contract permits arbitrary contract calls to be made with arbitrary payloads, opening up multiple attack vectors that can ultimately lead to a compromise of protocol funds.

As an example, any Boson voucher that is owned by the protocol itself can be exploited and acquired by an arbitrary party for free via the [`PriceDiscoveryBase::handleWrapper`](https://github.com/bosonprotocol/boson-protocol-contracts/blob/b29767e7be7b14ff87e69fb2709b4324956b93df/contracts/protocol/bases/PriceDiscoveryBase.sol#L252-L294) execution path by setting the `_priceDiscovery.priceDiscoveryContract` as the Boson Protocol Diamond and the `_priceDiscovery.priceDiscoveryData` payload as a `transferFrom` or `safeTransferFrom` [EIP-721](https://eips.ethereum.org/EIPS/eip-721) function.

### Impact:

Arbitrary calls to arbitrary contracts should not be allowed by any contract as they pose a significant security risk and can be exploited to siphon funds from the protocol. In the case of the Boson Protocol, this may be possible if an [EIP-20](https://eips.ethereum.org/EIPS/eip-20) asset supports a transfer-and-call flow (i.e. [ERC-677](https://github.com/ethereum/EIPs/issues/677)) or is an [EIP-721](https://eips.ethereum.org/EIPS/eip-721) / [EIP-1155](https://eips.ethereum.org/EIPS/eip-1155) asset.

### Example:

```solidity
function handleWrapper(
    uint256 _tokenId,
    address _exchangeToken,
    PriceDiscovery calldata _priceDiscovery,
    IBosonVoucher _bosonVoucher
) internal returns (uint256 actualPrice) {
    if (_tokenId == 0) revert TokenIdMandatory();

    // If price discovery contract does not own the voucher, it cannot be classified as a wrapper
    address owner = _bosonVoucher.ownerOf(_tokenId);
    if (owner != _priceDiscovery.priceDiscoveryContract) revert NotVoucherHolder();

    // Check balance before calling wrapper
    bool isNative = _exchangeToken == address(0);
    if (isNative) _exchangeToken = address(wNative);
    uint256 protocolBalanceBefore = getBalance(_exchangeToken, address(this));

    // Track native balance just in case if seller sends some native currency.
    // All native currency is forwarded to the wrapper, which should not return any back.
    // If it does, we revert later in the code.
    uint256 protocolNativeBalanceBefore = getBalance(address(0), address(this)) - msg.value;

    // Call the price discovery contract
    _priceDiscovery.priceDiscoveryContract.functionCallWithValue(_priceDiscovery.priceDiscoveryData, msg.value);

    // Check the native balance and revert if there is a surplus
    uint256 protocolNativeBalanceAfter = getBalance(address(0), address(this));
    require(protocolNativeBalanceAfter == protocolNativeBalanceBefore);

    // Check balance after the price discovery call
    uint256 protocolBalanceAfter = getBalance(_exchangeToken, address(this));

    // Verify that actual price is within the expected range
    if (protocolBalanceAfter < protocolBalanceBefore) revert NegativePriceNotAllowed();
    actualPrice = protocolBalanceAfter - protocolBalanceBefore;

    // when working with wrappers, price is already known, so the caller should set it exactly
    // If protocol receive more than expected, it does not return the surplus to the caller
    if (actualPrice != _priceDiscovery.price) revert PriceTooLow();

    // Verify that token id provided by caller matches the token id that the price discovery contract has sent to buyer
    getAndVerifyTokenId(_tokenId);
}
```

### Recommendation:

We advise the Boson Protocol to maintain a set of whitelisted `priceDiscoveryContract` entries, sanitizing the `_priceDiscovery.priceDiscoveryContract` entries. As an additional security measure, we advise the signatures permitted to be whitelisted as well by validating the first four bytes of the `_priceDiscovery::priceDiscoveryData` payload.

zajck:
Fix #861 

This change moves a part of price discovery logic outside the diamond, effectively creating another boson protocol client.

Since this client does not own any assets, they cannot be stolen from it. The user will always approve only Boson Protocol or external price discovery contract to transfer the vouchers and/or erc20 tokens.

Users can still make arbitrary calls through this client, but since it's outside the boson protocol address, any calls that the client makes will not be directly associated with the boson protocol.

### DIFF
--- a/contracts/protocol/bases/PriceDiscoveryBase.sol
+++ b/contracts/protocol/bases/PriceDiscoveryBase.sol
@@ -10,8 +10,6 @@ import { IBosonPriceDiscovery } from "../../interfaces/clients/IBosonPriceDiscov
 import { ProtocolBase } from "./../bases/ProtocolBase.sol";
 import { FundsLib } from "../libs/FundsLib.sol";
 
-// import { BosonPriceDiscovery } from "./../clients/priceDiscovery/BosonPriceDiscovery.sol";
-
 /**
  * @title PriceDiscoveryBase
  *
@@ -22,7 +20,7 @@ contract PriceDiscoveryBase is ProtocolBase {
 
     /**
      * @notice
-     * For offers with native exchange token, it is expected the the price discovery contracts will
+     * For offers with native exchange token, it is expected that the price discovery contracts will
      * operate with wrapped native token. Set the address of the wrapped native token in the constructor.
      *
      * @param _wNative - the address of the wrapped native token
--- a/contracts/protocol/bases/PriceDiscoveryBase.sol
+++ b/contracts/protocol/bases/PriceDiscoveryBase.sol
@@ -29,7 +29,7 @@ contract PriceDiscoveryBase is ProtocolBase {
     //solhint-disable-next-line
     constructor(address _wNative, address _bosonPriceDiscovery) {
         wNative = IWrappedNative(_wNative);
-        bosonPriceDiscovery = BosonPriceDiscovery(_bosonPriceDiscovery);
+        bosonPriceDiscovery = BosonPriceDiscovery(payable(_bosonPriceDiscovery));
     }
 
     /**
--- a/contracts/protocol/bases/PriceDiscoveryBase.sol
+++ b/contracts/protocol/bases/PriceDiscoveryBase.sol
@@ -118,15 +118,21 @@ contract PriceDiscoveryBase is ProtocolBase {
         );
         // ^^ we could skip 1 transfer if the caller approved bosonPriceDiscovery directly
 
-        return
-            bosonPriceDiscovery.fulfilAskOrder(
-                _tokenId,
-                _exchangeToken,
-                _priceDiscovery,
-                _buyer,
-                _bosonVoucher,
-                payable(msgSender())
-            );
+        actualPrice = bosonPriceDiscovery.fulfilAskOrder(
+            _exchangeToken,
+            _priceDiscovery,
+            _bosonVoucher,
+            payable(msgSender())
+        );
+
+        _tokenId = getAndVerifyTokenId(_tokenId);
+        {
+            // Make sure that the price discovery contract has transferred the voucher to the protocol
+            if (_bosonVoucher.ownerOf(_tokenId) != address(bosonPriceDiscovery)) revert VoucherNotReceived();
+
+            // Transfer voucher to buyer
+            _bosonVoucher.safeTransferFrom(address(bosonPriceDiscovery), _buyer, _tokenId);
+        }
     }
 
     /**
@@ -143,7 +149,6 @@ contract PriceDiscoveryBase is ProtocolBase {
      * - New voucher owner is not buyer wallet
      * - Token id sent to buyer and token id set by the caller don't match
      *
-     * @param _tokenId - the id of the token
      * @param _exchangeToken - the address of the exchange token
      * @param _priceDiscovery - the fully populated BosonTypes.PriceDiscovery struct
      * @param _seller - the seller's address
@@ -205,16 +210,6 @@ contract PriceDiscoveryBase is ProtocolBase {
 
         // Verify that token id provided by caller matches the token id that the price discovery contract has sent to buyer
         getAndVerifyTokenId(_tokenId);
-    }
-
-    /**
-     * @notice Returns the balance of the protocol for the given token address
-     *
-     * @param _tokenAddress - the address of the token to check the balance for
-     * @return balance - the balance of the protocol for the given token address
-     */
-    function getBalance(address _tokenAddress, address entity) internal view returns (uint256) {
-        return _tokenAddress == address(0) ? entity.balance : IERC20(_tokenAddress).balanceOf(entity);
     }
 
     /*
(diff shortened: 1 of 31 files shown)


contracts/protocol/bases/PriceDiscoveryBase.sol
```diff
@@ -6,20 +6,18 @@ import { ProtocolLib } from "../libs/ProtocolLib.sol";
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import { IWrappedNative } from "../../interfaces/IWrappedNative.sol";
 import { IBosonVoucher } from "../../interfaces/clients/IBosonVoucher.sol";
+import { IBosonPriceDiscovery } from "../../interfaces/clients/IBosonPriceDiscovery.sol";
 import { ProtocolBase } from "./../bases/ProtocolBase.sol";
 import { FundsLib } from "../libs/FundsLib.sol";
-import { Address } from "@openzeppelin/contracts/utils/Address.sol";
-import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+// import { BosonPriceDiscovery } from "./../clients/priceDiscovery/BosonPriceDiscovery.sol";
```
albertfolch-redeemeum, line 13: remove unused code

contracts/protocol/clients/priceDiscovery/BosonPriceDiscovery.sol
```diff
@@ -0,0 +1,275 @@
...
+import { FundsLib } from "../../libs/FundsLib.sol";
+import { Address } from "@openzeppelin/contracts/utils/Address.sol";
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import { BosonTypes } from "../../../domain/BosonTypes.sol";
+import { BosonErrors } from "../../../domain/BosonErrors.sol";
+
+/**
+ * @title BosonPriceDiscovery
+ *
+ * @dev Boson Price Discovery is a external contract that is used to determine the price of an exchange.
```
albertfolch-redeemeum, line 19:
```suggestion
 * @dev Boson Price Discovery is an external contract that is used to determine the price of an exchange.
```

contracts/protocol/clients/priceDiscovery/BosonPriceDiscovery.sol
```diff
@@ -0,0 +1,275 @@
...
+
+    IWrappedNative internal immutable wNative;
+
+    address private incomingTokenAddress;
+
+    address private immutable bosonProtocolAddress;
+
+    /**
+     * @notice
+     * For offers with native exchange token, it is expected the the price discovery contracts will
```
albertfolch-redeemeum, line 33:
```suggestion
     * For offers with native exchange token, it is expected that the price discovery contracts will
```

contracts/protocol/clients/priceDiscovery/BosonPriceDiscovery.sol
```diff
@@ -0,0 +1,275 @@
...
+     * operate with wrapped native token. Set the address of the wrapped native token in the constructor.
```
albertfolch-redeemeum, line 34:
```suggestion
     * operate with wrapped native token. Sets the address of the wrapped native token in the constructor.
```

contracts/protocol/clients/priceDiscovery/BosonPriceDiscovery.sol
zajck: Fixed.

contracts/protocol/clients/priceDiscovery/BosonPriceDiscovery.sol
zajck: Fixed.

contracts/protocol/clients/priceDiscovery/BosonPriceDiscovery.sol
zajck: I kept that as it was, since it's meant as an instruction for the deployer.

contracts/protocol/bases/PriceDiscoveryBase.sol
zajck: Removed

levalleux-ludo (dismissed):

albertfolch-redeemeum (approved):

levalleux-ludo (approved):

PR #904, opened 2024-02-14
Fix release funds

### LINKED ISSUE #902: Wrong price used in release funds
The release funds method always uses the original offer price:

https://github.com/bosonprotocol/boson-protocol-contracts/blob/9ed49d780f04b2f168001f72d6c09f5fc94818b7/contracts/protocol/libs/FundsLib.sol#L132

This is incorrect for offers with price discovery and offers with sequential commit since the last price is likely to be different from the starting price.

First the test suite should be updated and then the contracts should be fixed to release the correct amount.

zajck:
Fix #902

### DIFF
--- a/contracts/protocol/libs/FundsLib.sol
+++ b/contracts/protocol/libs/FundsLib.sol
@@ -9,6 +9,7 @@ import { ProtocolLib } from "../libs/ProtocolLib.sol";
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import { IBosonFundsLibEvents } from "../../interfaces/events/IBosonFundsEvents.sol";
+import { Math } from "@openzeppelin/contracts/utils/math/Math.sol";
 
 /**
  * @title FundsLib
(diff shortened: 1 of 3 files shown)


albertfolch-redeemeum (approved):

contracts/protocol/libs/FundsLib.sol
```diff
@@ -311,14 +322,17 @@ library FundsLib {
...
-                            ? effectivePriceMultiplier * (reducedSecondaryPrice - resellerBuyPrice)
-                            : (HUNDRED_PERCENT - effectivePriceMultiplier) * (resellerBuyPrice - reducedSecondaryPrice)
-                    ) /
-                    HUNDRED_PERCENT;
+                    applyPercent(resellerBuyPrice, (HUNDRED_PERCENT - effectivePriceMultiplier)) +
+                    secondaryCommit.price -
+                    applyPercent(secondaryCommit.price, (HUNDRED_PERCENT - effectivePriceMultiplier)) -
+                    applyPercent(secondaryCommit.protocolFeeAmount, effectivePriceMultiplier) -
+                    applyPercent(secondaryCommit.royaltyAmount, effectivePriceMultiplier) -
+                    Math.min(resellerBuyPrice, reducedSecondaryPrice);
```
levalleux-ludo, line 335:
```suggestion
                    applyPercent(resellerBuyPrice - secondaryCommit.price, (HUNDRED_PERCENT - effectivePriceMultiplier)) +
                    secondaryCommit.price -
                    applyPercent(secondaryCommit.protocolFeeAmount + secondaryCommit.royaltyAmount, effectivePriceMultiplier) -
                    Math.min(resellerBuyPrice, reducedSecondaryPrice);
```

levalleux-ludo (approved):
left 1 comment (possible gas optim?)

contracts/protocol/libs/FundsLib.sol
zajck: This cannot be changed for the following reasons:
1. `resellerBuyPrice - secondaryCommit.price` could be negative, so this would fail. We'd need to check if `resellerBuyPrice < secondaryCommit.price` and handle it differently.
2. `percent*A/100-percent*B/100` is not necessarily exactly the same as `percent*(A-B)/100`. For example, `A=5, B=3, percent=25` :
   1. `percent*A/100-percent*B/100=25*5/100+3*5/100=125/100-75/100=1-0=1`
   2. `percent*(A-B)/100=25*(5-3)/100=50/100=0`

levalleux-ludo: okay

PR #908, opened 2024-02-16
Secondary price should exceed cancellation penalty

### LINKED ISSUE #906: Make sure the secondary price is higher than cancelation penalty
If a buyer cancels the voucher, the last price must be high enough to cover the cancellation penalty, otherwise, the following line is broken.

https://github.com/bosonprotocol/boson-protocol-contracts/blob/7499ee78adf3f5def3d7c55e6a0445a65af2d55d/contracts/protocol/libs/FundsLib.sol#L153

This is relevant for the sequential commit only, since if the offer type is `discovery`, the cancellation penalty must be zero by default and any price is higher than that.

zajck:
Fix #906

### DIFF
--- a/contracts/interfaces/handlers/IBosonSequentialCommitHandler.sol
+++ b/contracts/interfaces/handlers/IBosonSequentialCommitHandler.sol
@@ -41,6 +41,7 @@ interface IBosonSequentialCommitHandler is BosonErrors, IBosonExchangeEvents, IB
      *   - Reseller did not approve protocol to transfer exchange token in escrow
      * - Call to price discovery contract fails
      * - Protocol fee and royalties combined exceed the secondary price
+     * - The secondary price cannot cover the buyer's cancelation penalty
      * - Transfer of exchange token fails
      *
      * @param _buyer - the buyer's address (caller can commit on behalf of a buyer)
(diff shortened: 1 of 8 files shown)


contracts/interfaces/handlers/IBosonSequentialCommitHandler.sol
```diff
@@ -41,6 +41,7 @@ interface IBosonSequentialCommitHandler is BosonErrors, IBosonExchangeEvents, IB
      *   - Reseller did not approve protocol to transfer exchange token in escrow
      * - Call to price discovery contract fails
      * - Protocol fee and royalties combined exceed the secondary price
+     * - The secondary price cannot cover the buyer's cancelation penalty
```
albertfolch-redeemeum, line 44:
```suggestion
     * - The secondary price cannot cover the buyer's cancellation penalty
```

contracts/protocol/facets/SequentialCommitHandlerFacet.sol
```diff
@@ -106,6 +105,9 @@ contract SequentialCommitHandlerFacet is IBosonSequentialCommitHandler, PriceDis
         // It might be lower than submitted for buy orders and higher for sell orders
         thisExchangeCost.price = fulfilOrder(_tokenId, offer, _priceDiscovery, seller, _buyer);
 
+        // Price must be high enough to cover cancelation penalty in case of buyer's cancelation
```
albertfolch-redeemeum, line 108:
```suggestion
        // Price must be high enough to cover cancellation penalty in case of buyer's cancelation
```

albertfolch-redeemeum (approved):

levalleux-ludo (approved):

PR #928, opened 2024-03-25
Prevent silent mint if the token exists already

### LINKED ISSUE #927: Prevent silentMint if token was transferred already
Before v2.4.0, `silentMint` was used to enable preminted vouchers. In those versions, transferring of preminted voucher always resulted in committing to the offer.
With v2.4.0 and price discovery, we allow for a single transfer of a preminted voucher which does not trigger the commit to offer. This is necessary for integration with price discovery mechanisms that require an actual transfer of the token (e.g. for on-chain auctions or AMMs).

The following code is now problematic:
https://github.com/bosonprotocol/boson-protocol-contracts/blob/d743c13d3458d6dd8b80ff4cd179bd5dcc28fca1/contracts/protocol/clients/voucher/BosonVoucher.sol#L775-L783

It allows the seller to transfer the voucher to the price discovery mechanism and then transfer it again even if it does not own it anymore. Although the seller cannot gain anything (assuming the price discovery mechanism is not malicious or buggy), it can result in locked (lost) buyer's funds.

To prevent it, make an additional check that `from` is the owner of the preminted voucher.

The bug cannot be exploited by a third-party (i.e. non-seller) account and was not present in earlier versions of the protocol.

zajck:
Fix #927


levalleux-ludo (approved):
LGTM

PR #975, opened 2025-01-23
Base sepolia addresses (v2.4.1)

zajck:
Protocol v2.4.1. was deployed to Base Sepolia. This PR includes addresses and ABIs for both testing and staging.

PR #976, opened 2025-01-23
Base sepolia addresses (v2.4.1)

zajck:
Protocol v2.4.1. was deployed to Base Sepolia. This PR includes addresses and ABIs for both testing and staging.

Replace #975 to include only the new changes.


albertfolch-redeemeum (approved):

levalleux-ludo (changes requested):
some addresses seem to be wrong in [scripts/config/protocol-parameters.js]

PR #979, opened 2025-02-03
Base addresses and hardhat upgrade

zajck:
- Added base contract addresses
- Upgraded hardhat to avoid avoid the problems with networks that do not have totalDifficulty field
- In new hardhat version, `selfdestruct` does not remove the contract code, which made some test outdated, so they were removed


zajck:
Superseded by #981

Issue #985, opened 2025-04-23
1. Protocol DR Fee Integration


0xlucian:
Implement core protocol changes to support DR fees.
• Remove fee amount validation that prevents non-zero fees in `DisputeResolverHandlerFacet.sol`
• Add fee collection during commit and refund during finalization
• Add fee handling during dispute escalation
• Add protocol-level mutualizer integration (call to `mutualizer` contract methods for fee coverage and returns)

Issue #984, opened 2025-04-22
3. Default Mutualizer + Minimal Mutualizer Interface


0xlucian:
This issue covers the implementation of:
	1.	A minimal mutualizer interface that external contracts must implement to be used in the protocol.
	2.	A default mutualizer contract that sellers can use if they don’t want to rely on third-party mutualizers.

**Minimal Mutualizer Interface**

Mutualizers are contracts that cover Dispute Resolution (DR) fees on behalf of sellers. The protocol only interacts with mutualizers through a defined interface.

Required methods:
	•	requestDRFee(seller, feeAmount, context) → (bool, uuid)
Called during commit. If the mutualizer accepts, it sends the DR fee to the protocol and returns a unique identifier.
	•	returnDRFee(uuid, feeAmount)
Called during exchange finalization. Indicates if the fee was used (feeAmount = 0) or returned.

Optional method:
	•	isSellerCovered(seller, feeAmount, requester, context) → bool
Read-only method to check if the seller is eligible for coverage. Useful for dApps to give early feedback to sellers.

**Default Mutualizer Contract**

A ready-to-use mutualizer contract that sellers can use directly.

Main features:
	•	Deposit and withdraw funds
	•	Create agreements between mutualizer and sellers (including limits, durations, and premiums)
	•	Sellers can activate coverage by paying a premium
	•	Mutualizer tracks agreement state and usage
	•	Agreements can be voided under certain conditions

Issue #983, opened 2025-04-22
2.Self-Mutualization (No external mutualizer)


0xlucian:
Allow sellers to pay DR fees directly from their pool.
	•	DRs can set non-zero fees
	•	DR fee is locked during commit and sent to the DR if the dispute is escalated and resolved
	•	Fee is returned if the dispute is not raised or is unresolved/refused

PR #1001, opened 2025-06-11
982 DR- Dispute Resolution System

### LINKED ISSUE #982: Dispute Resolution (DR) System
The protocol currently lacks a functional dispute resolution system. Dispute Resolvers (DRs) can be appointed but cannot charge fees, and the protocol does not manage DR fees or resolve disputes meaningfully.

This issue proposes introducing a full DR system, including:
	•	The ability for DRs to set and receive fees
	•	A mechanism to fund those fees using mutualizers
	•	Support for sellers to self-mutualize
	•	Protocol changes to pull/return DR fees
	•	A default mutualizer contract for sellers who don’t want to use third-party ones

0xlucian:
close #982 with subtasks: #983  #984 #985 

NOTE: branch - `1014-fundshandler-tests-refactor` has been merged into this one so the PR could be finalised.
Remember we first should merge the refactoring [PR](https://github.com/bosonprotocol/boson-protocol-contracts/pull/1029)

### DIFF
--- a/test/protocol/FundsHandlerTest.js
+++ b/test/protocol/FundsHandlerTest.js
@@ -1904,7 +1904,6 @@ describe("IBosonFundsHandler", function () {
       );
       expect(disputeResolver.isValid()).is.true;
 
-
       //Create DisputeResolverFee array so offer creation will succeed
       DRFee = parseUnits("0.01", "ether").toString();
       disputeResolverFees = [


contracts/protocol/libs/FundsLib.sol
```diff
@@ -207,6 +208,8 @@ library FundsLib {
             payoff.protocol += sequentialProtocolFee;
         }
 
+        handleDRFeeReturn(_exchangeId, exchange, offer, exchangeToken);
```
zajck, line 211: This seems to handle only the cases when fee is returned. 

We also need to handle when DR gets the fee and increase it's available funds. Probably `BosonTypes.Payoff` can be extended to include DR's payoff to and follow the pattern for others.

contracts/protocol/clients/DRFeeMutualizer.sol
```diff
@@ -0,0 +1,564 @@
...
+            (agreement.startTime > 0 && agreement.refundOnCancel) &&
+            (agreement.startTime + agreement.timePeriod > block.timestamp)
+        ) {
+            unchecked {
+                uint256 remainingTime = agreement.startTime + agreement.timePeriod - block.timestamp;
+                // Calculate refund as: premium * (remaining time / total time)
+                // Example: If premium is 100, timePeriod is 12 months, and 3 months remain:
+                // refund = 100 * 3 / 12 = 100 * 0.25 = 25
+                uint256 refundAmount = (agreement.premium * remainingTime) / agreement.timePeriod;
+                address tokenAddress = agreement.tokenAddress;
```
zajck, line 435: This line can be moved into `if (refundAmount > 0) {...}`

contracts/protocol/facets/ExchangeHandlerFacet.sol
```diff
@@ -1459,4 +1475,63 @@ contract ExchangeHandlerFacet is DisputeBase, BuyerBase, IBosonExchangeHandler {
...
+            );
+
+            uint256 balanceAfter = FundsLib.getBalance(exchangeToken);
+
+            uint256 feeTransferred = balanceAfter - balanceBefore;
+
+            if (!success || feeTransferred != _drFeeAmount) {
+                revert BosonErrors.DRFeeMutualizerCannotProvideCoverage();
+            } else {
+                FundsLib.increaseAvailableFundsAndEmitEvent(
```
zajck, line 1519: This is actually not consistent with how we otherwise handle the funds in the protocol.
For the funds that are just put in the escrow, we don't track them explicitly (refer commitToOffer or escalateDispute), so no available funds should be increased.  
Adding funds to a protocol entity makes them withdrawable by the protocol admin, which is not ok. Also, the emitted event `FundsReleased` is misleading. 

So at this point, actually nothing has to be done. At most `FundsEncumbered` event should be emitted, but even that is not strictly necessary, since `DRFeeRequested` contains the relevant information.

contracts/protocol/facets/OfferHandlerFacet.sol
```diff
@@ -402,6 +409,30 @@ contract OfferHandlerFacet is IBosonOfferHandler, OfferBase {
...
+     *
+     * @param _offerId - the id of the offer to update
+     * @param _newMutualizer - the new mutualizer address (can be zero for self-mutualization)
+     */
+    function updateOfferMutualizerInternal(uint256 _offerId, address _newMutualizer) internal {
+        // Make sure the caller is the assistant, offer exists and is not voided
+        Offer storage offer = getValidOfferWithSellerCheck(_offerId);
+
+        DisputeResolutionTerms storage disputeResolutionTerms = fetchDisputeResolutionTerms(_offerId);
+        disputeResolutionTerms.mutualizerAddress = payable(_newMutualizer);
```
zajck, line 431: Make sure here that old mutualizer is not the same as the new one, to prevent no-op transactions.

contracts/protocol/libs/FundsLib.sol
```diff
@@ -186,6 +193,11 @@ library FundsLib {
 
                     payoff.buyer = payoff.buyer + applyPercent(lastPrice, dispute.buyerPercent);
                     payoff.seller = payoff.seller + offerPrice - applyPercent(offerPrice, dispute.buyerPercent);
+
+                    // DR is always paid for Decided, but for Resolved it is paid only if escalated
+                    if (disputeState == BosonTypes.DisputeState.Decided || buyerEscalationDeposit > 0) {
```
zajck, line 198: I think it's enough to check only that `buyerEscalationDeposit > 0`.

Technically, yes, it's possible that `buyerEscalationDeposit == 0` and `disputeState == BosonTypes.DisputeState.Decided`, but in that case `feeAmount` will also be 0. (`buyerEscalationDeposit` is calculated as some percentage of `feeAmount`)

contracts/protocol/libs/FundsLib.sol
```diff
@@ -217,14 +229,56 @@ library FundsLib {
...
+                        _exchangeId,
+                        offer.sellerId,
+                        exchangeToken,
+                        returnAmount,
+                        sender
+                    );
+                }
+            } else {
+                if (returnAmount > 0) {
+                    decreaseAvailableFunds(PROTOCOL_ENTITY_ID, exchangeToken, returnAmount);
```
zajck, line 267: In connection with my comment in exchangeHandler - if we are not increasing funds there, we also should not decrease theme here.

zajck (changes requested):
There are a few more cases I'd like to see them tested:
- OfferHandler: createOffer with non-zero mutualizer address
- OfferHandler: test updateOfferMutualizer
- ExchangeHandler: it's possible to commit to an offer with mutualizer
- ExchangeHandler: commit fails if mutualizer cannot provide DRFee
- ExchangeHandler: commit fails if mutualizer sends wrong DRFee amount
- FundsHandler: all tests related to releaseFunds() must be updated (yes, that's a lot). 
   - in "📋 FundsLib  Methods" context, in `beforeEach` set `DRFee` to non-zero value. 
   - for all final states, check that DR got what it was supposed to get (even if the expected value is 0). This should affect only the existing tests.
   - additionally, create new tests with self-mutualization, since it affects the seller's payoff. This should be done for all final states.

That should lead to full coverage. The only thing that is left is `_msgData()` in `DRFeeMutualizer`, but that can be ignored (as we ignore it for BosonVoucher). Just change ci to allow for more skipped statements.

contracts/interfaces/IDRFeeMutualizer.sol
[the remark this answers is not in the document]
zajck: And also add the information about emitted events to natspec.

contracts/protocol/facets/ExchangeHandlerFacet.sol
```diff
@@ -1458,4 +1474,55 @@ contract ExchangeHandlerFacet is DisputeBase, BuyerBase, IBosonExchangeHandler {
...
+            }
+        }
+
+        // Emit event for DR fee request
+        emit IBosonFundsBaseEvents.DRFeeRequested(
+            _exchangeId,
+            exchangeToken,
+            _drFeeAmount,
+            _disputeTerms.mutualizerAddress,
+            msg.sender
```
zajck, line 1525:
```suggestion
            _msgSender()
```

contracts/protocol/bases/FundsBase.sol
```diff
@@ -228,8 +229,8 @@ abstract contract FundsBase is Context {
         }
 
         if (payoff.protocol > 0) {
-            increaseAvailableFunds(PROTOCOL_ENTITY_ID, exchangeToken, payoff.protocol);
-            emit IBosonFundsLibEvents.ProtocolFeeCollected(_exchangeId, exchangeToken, payoff.protocol, sender);
+            increaseAvailableFunds(0, exchangeToken, payoff.protocol);
```
zajck, line 232:
```suggestion
            increaseAvailableFunds(PROTOCOL_ENTITY_ID, exchangeToken, payoff.protocol);
```
I looks like `PROTOCOL_ENTITY_ID` got lost in one of the merges.I think it should be reinstated.

contracts/protocol/bases/OfferBase.sol
```diff
@@ -193,6 +195,19 @@ contract OfferBase is ProtocolBase, IBosonOfferEvents {
                             (feeAmount * fees.buyerEscalationDepositPercentage) /
                             HUNDRED_PERCENT;
                         disputeResolutionTerms.mutualizerAddress = _drParameters.mutualizerAddress;
+
+                        // Validate mutualizer interface if address is not zero (non-zero means external mutualizer)
+                        if (_drParameters.mutualizerAddress != address(0)) {
+                            try
```
zajck, line 201: I don't know if you considered what happens if `IERC165(_drParameters.mutualizerAddress).supportsInterface` returns something that cannot be decoded into a bool.  

I suppose it's still ok - the tx would fail (without a reason), but I don't see there's any attack there. Just the caller does not get the info why it failed. But maybe give antorher thought if you haven't already.

contracts/protocol/clients/DRFeeMutualizer.sol
```diff
@@ -0,0 +1,559 @@
...
+     * @dev Can be called by the seller or owner (if refundOnCancel is true). Calculates time-based refunds.
+     *
+     * Reverts if:
+     * - agreementId is invalid
+     * - Agreement is already voided
+     * - Caller is not authorized
+     * - ERC20 or native currency transfer fails
+     * - Seller not found
+     */
+    function voidAgreement(uint256 _agreementId) external {
```
zajck, line 391:
```suggestion
    function voidAgreement(uint256 _agreementId) external nonReentrant {
```
Probably it doesn't hurt to have guard here, since `IERC20(tokenAddress).safeApprove` is outside our control.

zajck (changes requested):

contracts/protocol/clients/DRFeeMutualizer.sol
```diff
@@ -0,0 +1,559 @@
...
+        if (_agreementId == 0 || _agreementId >= agreements.length) revert InvalidAgreementId();
+        bool premiumRefunded;
+        Agreement storage agreement = agreements[_agreementId];
+        if (agreement.isVoided) revert AgreementAlreadyVoided();
+
+        (bool exists, BosonTypes.Seller memory seller, ) = IBosonAccountHandler(BOSON_PROTOCOL).getSeller(
+            agreement.sellerId
+        );
+        if (!exists) revert SellerNotFound();
```
levalleux-ludo, line 400: This means it's not possible to void an agreement created for an non existing seller.
However, it is currently allowed to create an agreement for a non existing seller.
Not a major problem IMHO, but seems a bit inconsistent

contracts/protocol/clients/DRFeeMutualizer.sol
```diff
@@ -0,0 +1,559 @@
...
+        if (_maxAmountPerTx == 0) revert MaxAmountPerTxMustBeGreaterThanZero();
+        if (_maxAmountTotal < _maxAmountPerTx) revert MaxTotalMustBeGreaterThanOrEqualToMaxPerTx();
+        if (_timePeriod == 0) revert TimePeriodMustBeGreaterThanZero();
+
+        uint256 existingAgreementId = sellerToTokenToDisputeResolverToAgreement[_sellerId][_tokenAddress][
+            _disputeResolverId
+        ];
+        if (existingAgreementId != 0) {
+            Agreement storage existingAgreement = agreements[existingAgreementId];
+            if (
```
levalleux-ludo, line 348: If there is an already existing agreement for the same sellerId/tokenAddress/disputeResolferId, but not started yet (existingAgreement.startTime == 0), the method doesn't revert, creates a new agreement and replaces the agreementId in the sellerToTokenToDisputeResolverToAgreement map.
Not sure this behaviour is the one expected.
We could update the existing (not started) agreement with the new params, instead of created a new one.

levalleux-ludo: Maybe better => we can create a new agreement, after automatically voiding the previous one, so that we act its replacement with the new one.

contracts/protocol/clients/DRFeeMutualizer.sol
```diff
@@ -0,0 +1,559 @@
...
+                existingAgreement.startTime > 0 &&
+                !existingAgreement.isVoided &&
+                block.timestamp <= existingAgreement.startTime + existingAgreement.timePeriod
+            ) {
+                revert AgreementAlreadyExists();
+            }
```
levalleux-ludo, line 354: to follow the previous comment
```suggestion
            } else {
                agreement.isVoided = true;
                emit AgreementVoided(existingAgreementId, 0, 0);
            }
```

levalleux-ludo (changes requested):

contracts/protocol/bases/FundsBase.sol
```diff
@@ -217,14 +229,53 @@ abstract contract FundsBase is Context {
...
+                payoff.disputeResolver,
+                sender
+            );
+        }
+
+        // Return unused DR fee to mutualizer or seller's pool
+        if (drTerms.feeAmount != 0) {
+            uint256 returnAmount = drTerms.feeAmount - payoff.disputeResolver;
+
+            if (drTerms.mutualizerAddress == address(0)) {
```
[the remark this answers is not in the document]
zajck: That's correct, except we only need to add `mutualizerAddress` to exchange, while other DR terms can remain per offer only, since they cannot change.

contracts/protocol/clients/DRFeeMutualizer.sol
zajck: That's a good observation. Just my 2c:
- we don't want to update the existing agreement, since that would open up front-running attacks
- the real problem occurs if `newAgreement` and `payPremium` are racing. If `newAgreement` is processed 1s, the seller would pay for the old agreement which is not referenced in `sellerToTokenToDisputeResolverToAgreement` anymore.
- your 2nd suggestion seems ok. If `newAgreement` and `payPremium` are racing, whichever will be processed 2nd will revert.

test/protocol/FundsHandlerTest.js
```diff
@@ -2417,7 +2481,7 @@ describe("IBosonFundsHandler", function () {
           DRFee = parseUnits("0", "ether").toString();
```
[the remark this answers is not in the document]
zajck: Sorry, I commented this on the wrong line. I mean, in this test, it doesn't really matter, but the line where it's crucial is here:
https://github.com/bosonprotocol/boson-protocol-contracts/blob/208b68de43071865004a6d35087fc600b6ae0209/test/protocol/FundsHandlerTest.js#L1901

zajck (changes requested):
In my last review I already made this same comment, but I added it to the wrong line. The `DRFee` should be non zero here, in the setup of "📋 FundsBase  Methods" context (cannot comment there directly since it's not part of the diff):
https://github.com/bosonprotocol/boson-protocol-contracts/blob/208b68de43071865004a6d35087fc600b6ae0209/test/protocol/FundsHandlerTest.js#L1901

zajck (approved):
LGTM :medal_sports:

levalleux-ludo (approved):

PR #1029, opened 2025-07-09
Refactor FundsHandlerTest suite

### LINKED ISSUE #1014: Refactor FundHandlerTest suite
Current state of `FundHandlerTest` is a bit messy since there is alot of duplicated code which makes the whole test suite quite big and hard to maintain.

The main purpose of the refactoring is to make the code easier to maintain by lowering the number of repetition of same tests for different type of agent, mutualizer configurations, by just extracting those key configurations and cases in a list and just iterate over them by repeating same tests.

**Current areas of refactor:**

- **Agent fee configurations**: Multiple "Offer has an agent" contexts with duplicated test logic for different agent fee percentages

- **Final state scenarios**: Multiple different final states (COMPLETED, REVOKED, CANCELED, DISPUTED-*, etc.) repeated across contexts with identical test patterns

- **Sequential commit scenarios**: Multiple buyer chains and reseller scenarios (already partially refactored with configuration arrays) - demonstrates the target refactoring pattern

- **Protocol fee combinations**: Multiple fee percentage combinations tested in separate contexts

- **Token type variations**: ERC20, Native currency, and Price Discovery token tests with similar verification patterns

Example of similar refactoring that was done but never pushed can be seen [here](https://github.com/bosonprotocol/boson-protocol-contracts/blob/185d09ac7b38d3559af311dfaf9cdff5e385433d/test/protocol/FundsHandlerTest.js#L2332).

0xlucian:
close #1014

### DIFF
--- a/test/protocol/FundsHandlerTest.js
+++ b/test/protocol/FundsHandlerTest.js
@@ -2492,7 +2492,7 @@ describe("IBosonFundsHandler", function () {
           // Get the current block timestamp as dispute timestamp (just raised in parent)
           const currentBlock = await provider.getBlock("latest");
           const disputedDate = currentBlock.timestamp;
-          const timeout = BigInt(disputedDate) + BigInt(resolutionPeriod);
+          const timeout = disputedDate + Number(resolutionPeriod);
 
           // Set time to expire the dispute
           await setNextBlockTimestamp(Number(timeout) + 1);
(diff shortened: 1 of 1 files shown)


zajck (changes requested):
I suggested adding of some additional validations, but otherwise looks good.

Also, `applyPercentage` could be used in some places instead of writing it the long way.

zajck (approved):
LGTM

levalleux-ludo (approved):
approved but left a few suggestions (minor)

zajck (approved):

PR #1036, opened 2025-08-04
Create offer and commit in single tx

### LINKED ISSUE #1019: Create offer + commit to offer at the same time
BPIP: https://github.com/bosonprotocol/BPIPs/pull/37

Currently, the exchange starts on-chain, by offer creation. This makes the offer visible to buyers, and anyone can commit to it:
If it’s a fixed price offer, the buyer can do it without restrictions
If it’s a price discovery offer, it depends on the discovery mechanism. In some cases, the buyer can buy it immediately, but in some cases (e.g. if the item is on auction), the seller must explicitly accept the bid, so the seller can choose who they are selling to.  

The flow includes two main steps - “create offer” and “commit to the offer”. Even if the seller and the buyer decide on an exchange off-chain, they must still perform these two steps.

The protocol could be changed to also facilitate the exchanges where buyers and sellers fully agree on exchange terms before any information is stored in the protocol. Any of the involved parties could submit an agreement to the protocol, which would result in the same state as if one created an offer and the other committed to it.
The protocol role in this scenario is to validate the agreement, and do all the other actions that happen in the protocol after “commit to offer”, for example, Boson issues a tradeable rNFT, encumbers funds until the exchange finalisation and handles the disputes.  

The protocol's guarantees are preserved, since the buyer still gets the guarantee to either get the item or the money back. The only part that is left out compared to the existing protocol is the discovery phase.

Image

zajck:
Close #1019


0xlucian (changes requested):

contracts/protocol/facets/ExchangeCommitFacet.sol
```diff
@@ -216,14 +192,193 @@ contract ExchangeCommitFacet is DisputeBase, BuyerBase, OfferBase, IBosonExchang
...
+                uint256 buyerId = getValidBuyer(payable(_offerCreator));
+                if (buyerId != _fullOffer.offer.buyerId) {
+                    revert NotBuyerWallet();
+                }
+            }
+            EIP712Lib.verify(_offerCreator, offerHash, _signature);
+        } else if (offerId == VOIDED_OFFER_ID) {
+            // Offer is voided
+            revert OfferHasBeenVoided();
+        }
```
levalleux-ludo, line 379: nothing in case the offer already exist?

levalleux-ludo (changes requested):

contracts/protocol/bases/OfferBase.sol
```diff
@@ -464,4 +499,63 @@ contract OfferBase is ProtocolBase, BuyerBase, IBosonOfferEvents {
...
         if (totalRoyalties > _limits.maxRoyaltyPercentage) revert InvalidRoyaltyPercentage();
     }
+
+    /**
+     * @notice Computes the EIP712 hash of the full offer parameters.
+     *
+     * @param _fullOffer - the fully populated struct containing offer, offer dates, offer durations, dispute resolution parameters, condition, agent id and fee limit
+     * @return - the hash of the complete offer
+     */
+    function getOfferHash(BosonTypes.FullOffer calldata _fullOffer) internal view returns (bytes32) {
```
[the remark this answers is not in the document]
zajck: It might help a little, but not much.
1. For EOAs, any decent wallet will calculate the hash itself based on EIP712 spec. For users, it's important to understand what they are signing, and wallets expect only the type definition, domain parameters and input values.
2. For ERC1271, they can even fully ignore this hash if they want to. So, only if they decide to do the validation directly based on the hash, then it'll help them, but even then, there's still another hashing involved (done in `toTypedMessageHash`).

Anyway, since it's a small change, I implemented it.

contracts/protocol/facets/ExchangeCommitFacet.sol
zajck: We just continue, this is to enable offers `quantity > 1`.
If it exists, I must have been validated already, so there's no need to verify the signatures. It also means we can skip another creation of a duplicate offer.  

If there are any other invalid parameters (e.g. offer expired, is sold out, etc), that's validated in the commit to offer step.

PR #1048, opened 2025-09-22
Emit `FundsDeposited` on incoming transfer during commit

### LINKED ISSUE #1047: Emit `FundsDeposited` on every incoming payment
In certain cases, the funds are sent to the protocol without `FundsDeposited` being emitted.  
In those cases, other events (such as `FundsEncumbered`) are used to implicitly understand how much funds were transferred (and locked).  

This can lead to some inconsistencies, since it may look that more funds are locked than were ever deposited. To fix that, emit `FundsDeposited` every time `validateIncomingPayment` is invoked.

zajck:
Close #1047


levalleux-ludo (changes requested):
FundsEncumbered is also emitted from:
- escalateDisputeInternal()
- commitToPriceDiscoveryOffer()
- sequentialCommitToOffer()
The FundsDeposited event should also be emitted in these usecases

levalleux-ludo (approved):

PR #1103, opened 2026-04-28
ERC-3009 + EIP-2612 authorized transfer forwarder

zajck:
## Summary

New `BosonAuthorizedTransferForwarder` client contract that lets a token holder execute `depositFunds` or `commitToOffer` in a single transaction by signing an off-chain authorization — no separate `approve` tx required. Two flavours are supported:

- **ERC-3009** `receiveWithAuthorization` (USDC and other regulated stablecoins)
- **EIP-2612** `permit` (most modern DeFi tokens)

Both flavours additionally require an **action signature** binding the protocol-call parameters (entityId, or committer + offerId) to the inner authorization. Without it, an observer could front-run the tx and re-route funds. Action sig is verified before any token movement so a bad sig costs the attacker only base gas.

`commitToBuyerOffer` is intentionally not supported because the protocol identifies the seller via `_msgSender()`, which would require the forwarder to be registered as each seller's assistant — not viable for a generic per-deployment courier.

## What's in the PR

- **New contract**: `contracts/protocol/clients/BosonAuthorizedTransferForwarder.sol` — stateless forwarder, inherits OZ `EIP712`. Two ERC-3009 entry points + two EIP-2612 entry points, each taking an action `Signature` calldata struct. Permit calls are wrapped in `try/catch` so a front-runner consuming the permit nonce doesn't DoS the forwarder (the resulting allowance is what matters).
- **New interface**: `contracts/interfaces/IERC3009.sol`.
- **New mocks**: `Foreign20WithAuthorization.sol` (ERC-3009) and `Foreign20WithPermit.sol` (OZ ERC20Permit) — test-only.
- **Tests**: `test/protocol/clients/BosonAuthorizedTransferForwarderTest.js` — 41 cases covering happy/sad paths for both flows, front-run resistance for both, permit DoS resilience, inner-authorization invariants, and allowance/balance hygiene.
- **Deploy script**: `scripts/deploy-authorized-transfer-forwarder.js` + new hardhat task `deploy-authorized-transfer-forwarder --network <net> --env <env>`. Reads the existing `addresses/<chainId>-<network>-<env>.json` for `ProtocolDiamond`, deploys the forwarder bound to it, and appends the entry to the same address file. Refuses to overwrite an existing entry.
- **Testnet deployment artifacts**: `BosonAuthorizedTransferForwarder` entries appended to amoy / sepolia / basesepolia / opsepolia / arbsepolia test address files.

## Notes for reviewers

- The action sig is bound to the inner authorization via `(v, r, s)` of the inner sig embedded in the EIP-712 typehash — this implicitly commits to the full inner message (token domain, value, nonce, etc.). No separate forwarder nonce needed; replay protection comes for free from ERC-3009's one-shot bytes32 nonce or EIP-2612's per-account counter.
- Cross-chain replay is prevented on both signatures via chainId-bound EIP-712 domains.
- ECDSA-only for action sigs (matches ERC-3009, which is EOA-only in spec).
- Native-currency offers/deposits are out of scope (forwarder is not payable).
- No existing protocol contracts are modified.

## Test plan

- [x] `npx hardhat compile` — clean
- [x] `npx hardhat test test/protocol/clients/BosonAuthorizedTransferForwarderTest.js` — 41 passing
- [x] `npm run check:contracts` and `npm run check:scripts` — lint clean
- [x] Smoke-tested deploy task on local hardhat
- [x] Deployed on five test networks (addresses recorded in `addresses/*-test.json`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


thecryptofruit (dismissed):

zajck:
Replaced by #1104

PR #1155, opened 2026-05-26
test(reentrancy): add global nonReentrant guard matrix test suite

zajck:
## Summary

- Adds a comprehensive reentrancy regression test that walks the full (FROM × TO) matrix: every untrusted-callback surface that hands control out of the diamond (ETH \`receive\`, malicious ERC20, ERC721/1155 receiver hooks, external price discovery, sequential commit, DR fee mutualizer) attempts to re-enter every \`nonReentrant\` state-modifying entry point (~50 TO targets across ~18 facets). Each attempt must be blocked.
- Introduces \`contracts/mock/MaliciousReentrant.sol\` (base + \`MaliciousPriceDiscovery\` + \`MaliciousMutualizer\`). The mock self-disarms after a single attempt, optionally bubbles up the inner revert, and emits a \`ReentryAttempted\` event capturing the inner selector for assertion.
- Verification strategy is split: \`ReentryAttempted\` events for outer-survives flows (A–D) and \`revertedWithCustomError(bosonErrors, "ReentrancyGuard")\` for bubble-up flows (E, E', F). Twin-transfer flows (C, D) additionally assert the expected \`TwinTransferFailed\` / \`DisputeRaised\` follow-up.
- Adds an \`npm run test:reentrancy\` script and a parallel \`reentrancy-test\` job in CI (alongside \`test\`, \`integration-test\`, \`coverage\`).
- If any future change drops \`nonReentrant\` from a state-modifying entry point, the relevant matrix cases fail fast with a clear assertion — serving as a permanent regression guard.

## Test plan

- [x] \`npx hardhat test test/reentrancy/ReentrancyGuardTest.js\` passes (630 cases, ~14 s locally)
- [x] \`npm run check:contracts\` passes
- [x] \`npm run check:scripts\` passes
- [x] CI: new \`reentrancy-test\` job is green on the PR

🤖 Generated with [Claude Code](https://claude.com/claude-code)


contracts/mock/MaliciousReentrant.sol
```diff
@@ -0,0 +1,264 @@
...
+        return true;
+    }
+
+    function finalizeExchange(uint256, uint256) external override {
+        _attack();
+    }
+
+    function supportsInterface(bytes4 interfaceId) external pure returns (bool) {
+        return interfaceId == type(IDRFeeMutualizer).interfaceId || interfaceId == type(IERC165).interfaceId;
+    }
```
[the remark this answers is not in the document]
zajck: `IDRFeeMutualizer` is an interface and the `override` keyword is not required when implementing a function declared in an interface.

test/util/reentrancy-targets.js
```diff
@@ -0,0 +1,268 @@
...
+  "updateDisputeResolver",
+  "optInToDisputeResolverUpdate",
+  "addFeesToDisputeResolver",
+  "removeFeesFromDisputeResolver",
+  "addSellersToAllowList",
+  "removeSellersFromAllowList",
+  // ExchangeCommitFacet
+  "commitToOffer",
+  "commitToBuyerOffer",
+  "commitToConditionalOffer",
```
levalleux-ludo: missing `createOfferAndCommit` ?

zajck: I changed the file to programmatically get all functions, so `createOfferAndCommit` is now covered, too.

thecryptofruit (dismissed):

test/reentrancy/ReentrancyGuardTest.js
```diff
@@ -0,0 +1,1066 @@
...
+
+      snapshotA = await getSnapshot();
+    });
+
+    afterEach(async function () {
+      await revertToSnapshot(snapshotA);
+      snapshotA = await getSnapshot();
+    });
+
+    for (const to of buildReentrancyTargets()) {
```
levalleux-ludo: The matrix can silently provide ZERO coverage and still report green. [test/reentrancy/ReentrancyGuardTest.js:217](vscode-webview://0emrsu10houma20ankueubjcougfeaq6bt25jur1o5hrg3k14opr/test/reentrancy/ReentrancyGuardTest.js#L217), [test/util/reentrancy-targets.js:280](vscode-webview://0emrsu10houma20ankueubjcougfeaq6bt25jur1o5hrg3k14opr/test/util/reentrancy-targets.js#L280)
Each FROM block builds its its by iterating buildReentrancyTargets() in the describe callback — i.e. at mocha collection time, before any before() runs. The safety net expect(TO_TARGETS.length).to.be.greaterThan(0) lives in the outer before(). If buildReentrancyTargets() ever returns [] (facets present but no names matched), zero its are registered, so the outer before() never executes and the suite reports "0 passing" — the assertion is dead code that cannot catch the failure it was written for. The empty-return path is reachable: getMatrixTargetFunctionNames returns empty and the facet is continue-skipped with no error recorded whenever sourceKey = contracts/protocol/facets/${dir} fails to match a buildInfo.output.sources key (a facet moved to a subdirectory, or a file whose name ≠ contract name). The t.error check in before() also misses these silent skips. Move the non-empty/expected-count assertion to module scope (or a real it) and record an error for skipped facets.

test/util/reentrancy-targets.js
```diff
@@ -0,0 +1,420 @@
...
+ * delegation to an inner `nonReentrant` function, or a manual
+ * `reentrancyStatus == ENTERED` check), so an unguarded function shows up as
+ * a failure rather than silently disappearing from the matrix.
+ *
+ * @param {object} buildInfo - Hardhat build-info JSON
+ * @param {string} sourceKey - source path key inside buildInfo.output.sources
+ * @param {string} facetName - the facet's contract name
+ * @returns {Set<string>}
+ */
+function getMatrixTargetFunctionNames(buildInfo, sourceKey, facetName) {
```
levalleux-ludo: AST-based enumeration is the wrong depth: more complex than the ABI it already loads, and silently misses inherited functions. [test/util/reentrancy-targets.js:282](vscode-webview://0emrsu10houma20ankueubjcougfeaq6bt25jur1o5hrg3k14opr/test/util/reentrancy-targets.js#L282), [test/util/reentrancy-targets.js:163](vscode-webview://0emrsu10houma20ankueubjcougfeaq6bt25jur1o5hrg3k14opr/test/util/reentrancy-targets.js#L163)
buildReentrancyTargets loads both the artifact ABI (new Interface(artifact.abi)) and the 51 MB source build-info AST, then intersects them. But the facet artifact ABI already lists exactly the external/public functions — including those inherited from base contracts — with stateMutability. The AST contributes only the external-vs-public distinction, which is irrelevant for reentrancy. So the AST path is both heavier and less correct: any external/public state-modifying function inherited from a base sits in the ABI but is absent from the facet's own ContractDefinition AST nodes, so it's silently dropped from the matrix. (No current facet inherits such a function, so this is latent.) This also duplicates [scripts/util/diamond-utils.js:146](vscode-webview://0emrsu10houma20ankueubjcougfeaq6bt25jur1o5hrg3k14opr/scripts/util/diamond-utils.js#L146) (getStateModifyingFunctions) / [scripts/util/diamond-utils.js:17](vscode-webview://0emrsu10houma20ankueubjcougfeaq6bt25jur1o5hrg3k14opr/scripts/util/diamond-utils.js#L17) (getSelectors). Filtering the ABI directly (drop view/pure, drop isReentryAllowed) yields the same set, more completely. (The fully-correct source is the live diamond's DiamondLoupe selectors, but ABI-only is already a strict improvement.)

contracts/mock/MaliciousReentrant.sol
```diff
@@ -0,0 +1,264 @@
...
+
+/**
+ * @title MaliciousMutualizer
+ *
+ * @notice Variant of MaliciousReentrant that satisfies IDRFeeMutualizer so it
+ * can be registered as an offer's mutualizer. Every interface method routes
+ * through `_attack` to trigger re-entry into the protocol while the protocol
+ * is mid-flight in commit/finalize logic.
+ */
+contract MaliciousMutualizer is MaliciousReentrant, IDRFeeMutualizer {
```
levalleux-ludo: MaliciousMutualizer name collision (Hardhat HH701). [contracts/mock/MaliciousReentrant.sol:234](vscode-webview://0emrsu10houma20ankueubjcougfeaq6bt25jur1o5hrg3k14opr/contracts/mock/MaliciousReentrant.sol#L234)
A second contract named MaliciousMutualizer now exists alongside [contracts/mock/MockMutualizer.sol:64](vscode-webview://0emrsu10houma20ankueubjcougfeaq6bt25jur1o5hrg3k14opr/contracts/mock/MockMutualizer.sol#L64). This already forced the qualified-name edit in [test/protocol/ExchangeHandlerTest.js](vscode-webview://0emrsu10houma20ankueubjcougfeaq6bt25jur1o5hrg3k14opr/test/protocol/ExchangeHandlerTest.js) and the fully-qualified deploy string in the new test. Any future unqualified getContractFactory("MaliciousMutualizer") throws HH701. Renaming the new one (e.g. ReentrantMutualizer) removes the collision and the fragile qualified strings. (No other currently-broken call sites — checked by Claude AI.)

contracts/mock/MaliciousReentrant.sol
```diff
@@ -0,0 +1,264 @@
...
+                revert(add(ret, 0x20), mload(ret))
+            }
+        }
+        return ret;
+    }
+
+    /**
+     * @notice Read the selector encoded in the currently armed calldata.
+     */
+    function targetSelector() external view returns (bytes4) {
```
levalleux-ludo: Dead code: targetSelector(). [contracts/mock/MaliciousReentrant.sol:104](vscode-webview://0emrsu10houma20ankueubjcougfeaq6bt25jur1o5hrg3k14opr/contracts/mock/MaliciousReentrant.sol#L104) — never referenced in test/; it also duplicates the 4-byte extraction already inlined in _attack(). Remove it.

test/reentrancy/ReentrancyGuardTest.js
```diff
@@ -0,0 +1,1066 @@
...
+      it(`blocks reentry into ${to.name}`, async function () {
+        // The available funds were reset by the snapshot revert; redeposit
+        await fundsHandler.connect(rando).depositFunds(sellerId, ZeroAddress, depositAmount, { value: depositAmount });
```
levalleux-ludo: FROM A's per-test redeposit is redundant and its comment is wrong. [test/reentrancy/ReentrancyGuardTest.js:220](vscode-webview://0emrsu10houma20ankueubjcougfeaq6bt25jur1o5hrg3k14opr/test/reentrancy/ReentrancyGuardTest.js#L220) — snapshotA is taken after the before() deposit, so the revert restores funded state; the comment "available funds were reset by the snapshot revert" is incorrect and the extra depositFunds just doubles the balance.

test/reentrancy/ReentrancyGuardTest.js
```diff
@@ -0,0 +1,1066 @@
+/**
```
levalleux-ludo, line 1: Heavy before() copy-paste across FROM blocks C/D/E/E'/F. [test/reentrancy/ReentrancyGuardTest.js](vscode-webview://0emrsu10houma20ankueubjcougfeaq6bt25jur1o5hrg3k14opr/test/reentrancy/ReentrancyGuardTest.js) — the deploy-malicious + 3×grantRole + create-seller + create-DR + create-offer sequence is duplicated ~5×, and the offer-override lists already diverge subtly between blocks. A shared helper would cut ~100 lines and the drift risk.

levalleux-ludo (changes requested):

contracts/mock/MaliciousReentrant.sol
zajck: Removed

contracts/mock/MaliciousReentrant.sol
zajck: Renamed

test/reentrancy/ReentrancyGuardTest.js
zajck: Removed

test/util/reentrancy-targets.js
zajck: Now using ABI definition instead of AST.

test/reentrancy/ReentrancyGuardTest.js
zajck: Refactored

test/reentrancy/ReentrancyGuardTest.js
zajck: Changed `buildReentrancyTargets` to throw if no contracts found. Also `TO_TARGETS` are now built before describe so mocha uses the same set in every test.

levalleux-ludo (approved):

thecryptofruit (approved):

